PR #1, opened 2017-05-19
box template: added mock environment for build

tomaskrizek:
Box template is now generated with mock environment with the
necessary build dependencies. An ansible playbook can be used
to build rpms from git sources.


tomaskrizek:
Tested generating template box with this patch. The ansible build passes.

PR #2, opened 2017-05-24
ansible: fix build artifacts and organize roles

tomaskrizek:
Box template had artifacts from build (`/root/rpmbuild`, `/root/freeipa`). I re-organized the roles to allow proper cleanup after the build.

To keep the code readable and consistent, I also re-organized other roles and tasks to follow the same directory structure.


tomaskrizek:
Artifacts are absent in the new box template.

PR #3, opened 2017-05-25
template: add workaround for gssproxy nfs-utils issue

tomaskrizek:
systemctl restart gssproxy would fail during ipa-server-install,
becase nfs-utils was installed, but nfs wasn't started.

Remove once https://bugzilla.redhat.com/show_bug.cgi?id=1452881
is fixed.


tomaskrizek:
Tested the new template. I didn't hit the gssproxy issue during ipa-server-install.

PR #29, opened 2017-08-01
github: Use per runner instance of redis for caching

dkupka:


tomaskrizek:
Works, reduced amount of requests from ~100 per reboot to about 10. 

Could please also link to the docs where you found how often the save happens?

davidkupka:
From self documented configuration file (https://raw.githubusercontent.com/antirez/redis/2.4/redis.conf):

```
################################ SNAPSHOTTING  #################################
#
# Save the DB on disk:
#
#   save <seconds> <changes>
#
#   Will save the DB if both the given number of seconds and the given
#   number of write operations against the DB occurred.
#
#   In the example below the behaviour will be to save:
#   after 900 sec (15 min) if at least 1 key changed
#   after 300 sec (5 min) if at least 10 keys changed
#   after 60 sec if at least 10000 keys changed
#
#   Note: you can disable saving at all commenting all the "save" lines.

save 900 1
save 300 10
save 60 10000
```

PR #34, opened 2017-08-03
github: Log when GitHub API request is consumed

davidkupka:


tomaskrizek:
Do we really want this? I'm afraid it might pollute the logs unnecessarily... Perhaps there could be an option to turn this on during debugging and have it off by default?

PR #37, opened 2017-08-04
[fix #20] Handling provision issues making it run twice.

### LINKED ISSUE #20: Handle provisioning issues
Occasionally, some one-time provisioning issue is encountered during vagrant up, such as

- [libvirt](https://fedorapeople.org/groups/freeipa/prci/jobs/f4b7abc8-744a-11e7-b922-001a4a23162d/runner.log.gz): Call to virConnectOpen failed: Failed to connect socket to '/var/run/libvirt/libvirt-sock': No such file or directory
- [nfs](https://fedorapeople.org/groups/freeipa/prci/jobs/681c5ef8-73f9-11e7-9fca-001a4a23166c/runner.log.gz): /usr/bin/mv: cannot create regular file '/etc/exports': File exists
- [hostnamectl](https://fedorapeople.org/groups/freeipa/prci/jobs/2b2274f6-743f-11e7-9fca-001a4a23166c/runner.log.gz): Could not set property: Failed to activate service 'org.freedesktop.hostname1': timed out (service_start_timeout=25000ms)

These problems aren't persistent. When vagrant up or provision fails, cleanup should be performed and then another one or two attempts to provision the machines. If they all fail, then an error should be reported to github as it is now.

felipevolpone:
Also, fixing minor typo error.

Fixes: https://github.com/freeipa/freeipa-pr-ci/issues/20

### DIFF
--- a/tasks/tasks.py
+++ b/tasks/tasks.py
@@ -9,27 +9,7 @@ from .common import (FallibleTask, TaskException, PopenTask,
                      logging_init_file_handler, create_file_from_template)
 from . import constants
 from .remote_storage import GzipLogFiles, FedoraPeopleUpload
-from .vagrant import VagrantUp, VagrantProvision, VagrantCleanup
-
-
-def with_vagrant(func):
-    def wrapper(self, *args, **kwargs):
-        try:
-            self.execute_subtask(
-                VagrantUp(timeout=None))
-            self.execute_subtask(
-                VagrantProvision(timeout=None))
-        except TaskException as exc:
-            logging.critical('vagrant or provisioning failed')
-            raise exc
-        else:
-            func(self, *args, **kwargs)
-        finally:
-            if not self.no_destroy:
-                self.execute_subtask(
-                    VagrantCleanup(raise_on_err=False))
-
-    return wrapper
+from .vagrant import with_vagrant
 
 
 class JobTask(FallibleTask):
--- a/tasks/vagrant.py
+++ b/tasks/vagrant.py
@@ -5,6 +5,39 @@ import os
 from .common import PopenTask, PopenException, FallibleTask, TaskException
 
 
+def with_vagrant(func):
+    def wrapper(self, *args, **kwargs):
+        try:
+            __setup_provision(self)
+        except TaskException as exc:
+            logging.critical('vagrant or provisioning failed')
+            raise exc
+        else:
+            func(self, *args, **kwargs)
+        finally:
+            if not self.no_destroy:
+                self.execute_subtask(
+                    VagrantCleanup(raise_on_err=False))
+
+    return wrapper
+
+
+def __setup_provision(task):
+    """
+    This tries to execute the provision twice due to
+    problems described in issue #20
+    """
+    try:
+        task.execute_subtask(VagrantUp(timeout=None))
+        task.execute_subtask(VagrantProvision(timeout=None))
+    except Exception as exc:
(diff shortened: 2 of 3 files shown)


tasks/tasks.py
```diff
@@ -32,6 +29,23 @@ def wrapper(self, *args, **kwargs):
...
+    This tries to execute the provision twice due to
+    problems described in issue #20
+    """
+    vm_is_up = False
+    try:
+        self.execute_subtask(VagrantUp(timeout=None))
+        vm_is_up = True
+        self.execute_subtask(VagrantProvision(timeout=None))
+    except:
+        self.execute_subtask(VagrantCleanup(raise_on_err=False))
```
tomaskrizek: This will destroy the VMs, so `vm_is_up` should be omitted.

tasks/tasks.py
```diff
@@ -15,10 +15,7 @@
 def with_vagrant(func):
```
tomaskrizek: Since this function's scope is starting to grow, I'd prefer to move the whole `with_vagrant` decorator to `tasks/vagrant.py`

tasks/tasks.py
felipevolpone: you're right, my misunderstood of how VagrantCleanup works.

tasks/vagrant.py
```diff
@@ -5,6 +5,38 @@
...
+            func(self, *args, **kwargs)
+        finally:
+            if not self.no_destroy:
+                self.execute_subtask(
+                    VagrantCleanup(raise_on_err=False))
+
+    return wrapper
+
+
+def __setup_provision(self):
```
tomaskrizek: I don't mind the `self` usage in the decorator's `wrapper` function, but I'd rather avoid it here, because it's a separate function and it's a bit confusing to have the `self` keyword outside a class. `task` might be better here.

tasks/vagrant.py
```diff
@@ -5,6 +5,38 @@
...
+    """
+    This tries to execute the provision twice due to
+    problems described in issue #20
+    """
+    try:
+        self.execute_subtask(VagrantUp(timeout=None))
+        self.execute_subtask(VagrantProvision(timeout=None))
+    except:
+        logging.debug("Failed to provision/up VM. Trying it again")
```
tomaskrizek: `debug` is mostly used for stdout spam etc. I think this is worthy of `info` level.

tasks/vagrant.py
felipevolpone: Well, I though to change `self` to `instance` (or something like that, but I think it would even more confuse.

I think that, by the reason that use `self` outside of a class method is weird, we should keep the `self` in `__setup_provision`. IMHO it's more explicit.

tomaskrizek: I still think `task` would be better, but I leave it up to you.

tasks/vagrant.py
```diff
@@ -5,6 +5,38 @@
...
+
+def __setup_provision(task):
+    """
+    This tries to execute the provision twice due to
+    problems described in issue #20
+    """
+    try:
+        task.execute_subtask(VagrantUp(timeout=None))
+        task.execute_subtask(VagrantProvision(timeout=None))
+    except:
```
felipevolpone: @tomaskrizek not sure if you want to log the error here (as debug at least). up to you

tomaskrizek: Added the debug msg.

PR #38, opened 2017-08-05
Improving README with info about setting ssh connection in dev environment

felipevolpone:
Adding content explaining that the user will need to setup the ssh connection as root.

A preview can be checked here: https://github.com/felipevolpone/freeipa-pr-ci/tree/readme-ssh-conn

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ contributors, you can check out this [video](https://vimeo.com/228077191).
 
 2. Place `freeipa_pr_ci` private key in `keys/`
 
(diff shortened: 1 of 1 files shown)


tomaskrizek:
429954978c08b1dce6824104e6743643deb46667 addressed this, the playbooks will attempt to connect as root by default, there's no need to configure it manually now. The only thing a use may need to do is to set the proper ssh key to be used.

felipevolpone:
nice!

PR #66, opened 2017-08-24
runner: make sure prci is always enabled after deployment

tomaskrizek:


felipevolpone:
Tested and it works properly. ACK.
Just being curious: why are you changing from the handlers approach to adding more steps in the task?
TBH I prefer the approach that this patch brings, it's more straightforward IMHO.

tomaskrizek:
@felipevolpone The handler is only notified (and executed) if the task has changed. There were cases where `prci.service` file wasn't updated, and the playbook wouldn't turn the service on if it was previously turned off (for example manually). Adding it as a separate step ensures it's always executed.

Issue #54, opened 2017-08-17
runner: allow execution of multiple jobs in parallel


tomaskrizek:
A runner should be capable of executing multiple jobs. Initial implementation could have a predefined number of maximum concurrent jobs. In the future, each job could have the required resources associated with it. Then, the runner would be capable of fully utilizing its resources by executing as many jobs as it has resources for.

tomaskrizek:
We've tried to implement this functionality in #114, which was based on #72. The approach was to fork the process once the task queue was generated and execute the individual tasks in their own processes. This was more difficult than expected and led to many hard-to-debug issues with communication over ssl (https://stackoverflow.com/questions/3724900/python-ssl-problem-with-multiprocessing, https://stackoverflow.com/questions/25455462/share-list-between-process-in-python-server). To avoid breaking existing functionality, we've decided to postpone this until we can implement it properly.

Ideally, generating the task queue should be completely separate from the code that is capable of executing a single task. Then, some kind of messaging should be used to communicate between the processes (e.g. [RabbitMQ](http://rabbitmq.com/)). It should be implemented in a way that allows the task queue to be generated on a different machine than the machine that will be executing the tasks. This would also enable future scaling of the system.

tomaskrizek:
It might be possible to solve the multiprocessing issue with SSL:

```
Applications must change the PRNG state of the parent process if they use any
SSL feature with os.fork(). Any successful call of RAND_add(), RAND_bytes() 
or RAND_pseudo_bytes() is sufficient.
```

https://docs.python.org/3/library/ssl.html#multi-processing

f-trivino:
We have several testing scenarios (jobs) where memory consumption is close to the limit. This feature will cause multiple issues when running jobs. The implementation is difficult and the added value is not significant. Also, when using different templates we will not have enough hard disk quota. We close this ticket.

PR #72, opened 2017-08-30
[WIP] github: Execute tasks in parallel to fully utilize runner

davidkupka:
DON'T EXECUTE THIS. It's free-hand coding and it's guaranteed to misbehave.
This is my idea for executing tasks in parallel which would let us utilize beefy runners fully.

### DIFF
--- a/github/prci.py
+++ b/github/prci.py
@@ -24,42 +24,10 @@ SENTRY_URL = 'https://d24d8d622cbb4e2ea447c9a64f19b81a:4db0ce47706f435bb3f8a02a0
 ERROR_BACKOFF_TIME = 600
 REBOOT_DELAY = 3600 * 3
 REBOOT_TIME_FILE = '/root/next_reboot'
+REBOOT_CHECK_INTERVAL = 300
+TASK_MIN_CPU = 2
+TASK_MIN_MEM = 4*2**30
 logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
-
-
-class ExitHandler(object):
-    done = False
-    aborted = False
-    task = None
-
-    def finish(self, signum, frame):
-        if self.done:
-            return self.abort(signum, frame)
-
-        logger.info("Waiting for current task to finish. This may take long.")
-        self.done = True
-
-    def abort(self, signum, frame):
-        if self.aborted:
-            return self.quit()
-
-        if self.task:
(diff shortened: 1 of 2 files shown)


tomaskrizek:
The approach looks good. Are you aware of anything that's still missing?

davidkupka:
1. All jobs in .freeipa-pr-ci.yaml should be extended with
```
resources:
    cpu: N  # number of CPU cores needed by job
    memory: M  # amount of RAM needed by job (in bytes)
```
otherwise it's assumed that the job needs whole runner and only one task at the time is executed.

2. Handle termination better. Currently, first SIGINT sets Executive.done to True meaning that it won't start any new jobs and will wait for running ones to finish. Second SIGINT calls Executive.abort which sends SIGTERM to all child processes and wait for them to finish. Third SIGINT calls sys.exit. It sort of makes sense but only if subprocesses handle SIGTERM as a request to abort and clean up immediately.

3. TASK_MIN_CPU and TASK_MIN_MEM should be moved to configuration file. Those variables define the minimal amount of resources left on runner when it makes sense to look for jobs to execute.

github/prci.py
```diff
@@ -211,32 +179,109 @@ def sentry_report_exception(context=None):
...
+        self.done = True
+        self.reboot = False
+
+    def abort(self):
+        if self.abort:
+            sys.exit('Forced quit. There may be stale files and processes '
+                     'left behind.')
+        logging.debug('Abort. Waiting for clean up.')
+        self.done = True
+        self.abort = True
```
tomaskrizek: This should also call terminate() for all processes.

github/prci_github/internals.py
```diff
@@ -13,6 +13,7 @@
 import yaml
 
 import cachecontrol
+import psutil
```
tomaskrizek: Add psutil to ansible playbook dependencies.

felipevolpone:
@davidkupka thank you very much for the patch :)) I've merged into the code and changed a little if you don't mind.  PR #114 address it.

tomaskrizek:
We've decided to abandon this approach, as it proved to be very error prone and we don't want to break the existing infrastructure. See https://github.com/freeipa/freeipa-pr-ci/issues/54#issuecomment-344917531 for details.

PR #79, opened 2017-09-14
tasks: write hostname to file before job execution

tomaskrizek:
A hostname can be useful to identify which machine executed the job,
which can help to debug issues with a particular runner.


stlaz:
I don't see much harm in this. ACK.

PR #83, opened 2017-09-23
github: Regenerate only failed tasks when RERUN_LABEL appears

davidkupka:
!!!UNTESTED!!!
I don't have any PR CI deployment so this is untested.

### DIFF
--- a/github/prci_github/internals.py
+++ b/github/prci_github/internals.py
@@ -349,6 +349,17 @@ class Tasks(collections.Set, collections.Mapping):
                           'pending')
         logger.debug("Creating tasks for PR %d done.", self.pull.pull.number)
 
+    def create_missing(self):
+        logger.debug("Creating missing tasks for PR %d", self.pull.pull.number)
+        for task in self.tasks_conf:
+            try:
+                self[task]
+            except KeyError:
+                Status.create(self.repo, self.pull, task, 'unassigned', '',
+                              'pending')
+        logger.debug("Creating missing tasks for PR %d done.",
+                     self.pull.pull.number)
+
 
 class TaskQueue(collections.Iterator):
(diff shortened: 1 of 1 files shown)


tomaskrizek:
Thanks for the contribution, David! :) I hope to take a look at this PR as soon as I get back to work next week.

tomaskrizek:
I've made some minor adjustments (syntax fix, removing the re-run label), hopefully you don't mind :)

I also found out it slightly changes the behaviour of the `re-run` label. Consider a new test case is added to target branch after tasks have been generated. Before, the new test case could be triggered by adding `re-run` label. Would it be possible to keep this behaviour?

davidkupka:
@tomaskrizek I've updated the PR so it:
1) removes `re-run` label as soon as possible to minimize risk of race condition. The risk is still there but since the worse thing that may happen is multiple unnecessary recreation of statuses, I don't think we need to do anything about it. 
2) creates missing tasks as requested :-)

felipevolpone:
Tested and it works properly.
I just think the `create_tasks_for_pulls` it was getting too big, so I split it and made it more simple to read from my POV.

github/prci_github/internals.py
```diff
@@ -363,47 +374,66 @@ def create_tasks_for_pulls(self):
...
+                        Status.create(task.repo, task.pull, task.name,
+                                      'unassigned', '', 'pending')
+                # The most likely reason to create missing tasks is
+                # missing is that new task was added in master branch.
+                # Other reason is that someone removed that task manually.
+                tasks.create_missing()
+
+            self._rerun_staled_tasks(tasks)
+
+    def _rerun_staled_tasks(self, tasks):
```
tomaskrizek: typo, `stalled`

tomaskrizek:
Just a minor nitpick typo. If it works, feel free to push it.

PR #86, opened 2017-10-04
Adding dependencies to run ipa-tests with python3

felipevolpone:
It's necessary to use paramiko v2.2.1 due to issues with gssapi (https://github.com/paramiko/paramiko/issues/1069)

### DIFF
--- a/ansible/roles/machine/setup/tasks/install_packages.yml
+++ b/ansible/roles/machine/setup/tasks/install_packages.yml
@@ -36,12 +36,16 @@
     - PyYAML  # ipa-vagrant-ci dependency
     - vim
     - NetworkManager
-    - python3-paramiko
 
-- name: install pip dependencies
+- name: install py3 pip dependencies
   pip:
-    name: pytest-html
-    state: latest
+    executable: pip3
+    name: "{{ python_packages_to_install | join(' ') }}"
+
+- name: install py2 pip dependencies
+  pip:
+    executable: pip
+    name: "{{ python_packages_to_install | join(' ') }}"
 
 - name: disable distro repositories
   shell: "dnf config-manager --set-disabled {{ item }}"
--- a/ansible/roles/machine/setup/defaults/main.yml
+++ b/ansible/roles/machine/setup/defaults/main.yml
@@ -1,4 +1,8 @@
 ---
 selinux_mode: permissive
 copr_repo: "@freeipa/freeipa-master"
-
+python_packages_to_install:
+  - pytest-html
+  - paramiko==2.2.1
+# It's necessary to use paramiko v2.2.1 due to issues
+# with gssapi (github.com/paramiko/paramiko/issues/1069)


felipevolpone:
Box including this patch (https://app.vagrantup.com/freeipa/boxes/ci-master-f26/versions/0.1.6) released.

ansible/roles/machine/setup/tasks/install_packages.yml
```diff
@@ -36,13 +36,19 @@
...
   pip:
+    executable: pip3
     name: pytest-html
     state: latest
 
+- name: install paramiko
+  pip:
+    executable: pip3
+    name: paramiko
+    version: 2.2.1
```
tomaskrizek: Why do we require this specific version?

ansible/roles/machine/setup/tasks/install_packages.yml
```diff
@@ -36,13 +36,19 @@
     - PyYAML  # ipa-vagrant-ci dependency
     - vim
     - NetworkManager
-    - python3-paramiko
```
tomaskrizek: Is the package in fedora dependencies broken? Why is it installed with pip instead?

ansible/roles/machine/setup/tasks/install_packages.yml
```diff
@@ -36,13 +36,19 @@
...
 
 - name: install pip dependencies
   pip:
+    executable: pip3
```
tomaskrizek: This breaks backwards compatibility. Before, it was possible to run ipa tests with python2 -- if you don't install this package with pip as well, it won't be possible to run tests with python 2 anymore.

I'd either install this with both pip and pip3, or bump the template version to 0.2.x to signify some compatibility/API has been changed.

ansible/roles/machine/setup/tasks/install_packages.yml
tomaskrizek: I see you mentioned the issue in the commit message. Could you please link the issue with a comment in the code as well?

ansible/roles/machine/setup/tasks/install_packages.yml
felipevolpone: I see we are not on the same page here.
I thought we would have new templates only for python 3 stuff. I mean, we could stay with 0.1.4 to run tests with py2, and the new templates would be only to help PR 1112. Once all tests were migrated to py3, we would drop the support for 0.1.4.

However, if this is not the idea, I'll add the installation with `pip` as well.

ansible/roles/machine/setup/tasks/install_packages.yml
felipevolpone: Sure.

felipevolpone:
JFTR, if we do not use 2.2.1 we got these errors: https://pastebin.com/2kgghyxb

ansible/roles/machine/setup/tasks/install_packages.yml
tomaskrizek: The disk space on the runners is very limited (less than 20GB for templates, each template takes ~2.8GB). Keep in mind each branch can (and may have to) have a separate template. Also, we may want to test on multiple distros in the future. There should never be more than a single supported version for branch/distro, we simply don't have the space for it.

tomaskrizek: Also, any breaking changes or significant updates should bump either x or y in x.y.z version format. The default z bump is intended primarily for minor changes and package updates. Supporting py2 in 0.1.4 and not supporting it in 0.1.6 is not a good idea.

felipevolpone:
Requested changes done, if ack I can build the template again

tomaskrizek:
I've uploaded the new box: [0.1.7](https://app.vagrantup.com/freeipa/boxes/ci-master-f26/versions/0.1.7)

PR #89, opened 2017-10-06
Add a filter for hypervisor machines in the job xml template

apophys:


tomaskrizek:
Works fine, we just need #129 to be able to properly handle AMD machines.

PR #92, opened 2017-10-06
runner: don't wait for reboot after deployment

tomaskrizek:
Often the 2 minute timeout wasn't enough for the machine to come back
and the playbook would seemingly fail on the last step. Even though the
error was harmless, it could confuse users.

### DIFF
--- a/ansible/roles/runner/handlers/main.yml
+++ b/ansible/roles/runner/handlers/main.yml
@@ -10,11 +10,8 @@
   poll: 0
   ignore_errors: true
   notify:
-    - wait for machine
+    - reboot_message
 
-- name: wait for machine
-  local_action: wait_for
-    host="{{ inventory_hostname }}"
-    port=22
-    delay=10
-    timeout=120
+- name: reboot_message
+  debug:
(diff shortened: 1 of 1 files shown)


tomaskrizek:
This should be simple enough to review (just run the playbook on a new runner) and it'd be great if this would be reviewed before the workshop next Tuesday. Please take a look if you have the time.

ansible/roles/runner/handlers/main.yml
```diff
@@ -10,11 +10,15 @@
   poll: 0
   ignore_errors: true
   notify:
-    - wait for machine
+    - reboot_message
 
 - name: wait for machine
```
apophys: Is this handler used from somewhere else? If not, why do you keep it here?

tomaskrizek: It's not used anywhere else. It's just a quick way to run multiple tasks when I trigger a handler. Do you see a better place for the code?

felipevolpone:
Tested and it works properly.

PR #102, opened 2017-10-17
ansible: make callback plugin forward compatible

tomaskrizek:
Ansible 2.4.0 made some backward incompatible changes which prevented
human_log.py to work properly. This ensures it works with both <2.4 and
>= 2.4 Ansible version.

Issue in Ansible: https://github.com/ansible/ansible/issues/30597


abbra:
LGTM

PR #114, opened 2017-10-27
Allowing multiple topologies

### LINKED ISSUE #70: runner: support multiple topologies
Currently, only master/replica topology is supported for testing. It'd be ideal if we could define multiple topologies and specify which one should be used for a given job. The topologies could also have defined how many resources (CPU/RAM) they require and this information could be used to fully utilize runner's capacity (see #54)

felipevolpone:
In order to have more tests running in upstream PR CI
infrastructure, we need to support multiple topologies:
master with multiple replicas and clients. This PR address this.

To understand the resources that a given task should have to
run it, the freeipa/freeipa/.freeipa-pr-ci.yml file will
declare how much RAM and CPUs is needed. Using this syntax:

```
topologies:
  master_1repl: &master_1repl
    name: master_1repl
    cpu: 2
    memory: 4098
    name: master_1repl_1client
    cpu: 6
    memory: 4098
```

After declaring the topologies, the tasks will be able
to use it with this syntax:

```
fedora-26/simple_replication:
    ...
    job:
      class: RunPytest
      args:
        ...
        topology: *master_1repl
```

Is IMPORTANT to say that, the name of the topology
must match with the name of a Vagrantfile inside the
templates/vagrantfiles dir. So, if a topology is added to
the .freeipa-pr-ci.yml file, a correspoding Vagrantfile
must be added in the freeipa-pr-ci project as well.

The Task class will read the config file and the Scheduler
will deal with it, to know if the runner is able to run
that task.

Fixes #70


felipevolpone:
TASK_MIN_CPU and TASK_MIN_MEM could be moved to a configuration file

felipevolpone:
Once we have an agreement about this PR, I'll open a PR against freeipa adding the topologies in `.freeipa-pr-ci.yml` file.

tasks/tasks.py
```diff
@@ -185,28 +189,46 @@ class RunPytest(JobTask):
...
         self.build_url = build_url + '/'
         self.test_suite = test_suite
 
+        if not topology:
+            topology = {'name': constants.DEFAULT_TOPOLOGY}
+
+        self.topology_name = topology.get('name')
+
     def _before(self):
-        super(RunPytest, self)._before()
```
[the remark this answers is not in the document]
felipevolpone: If you call super in this case, it will try to create a vagrantfile using the action_name instead of topology_name (lines 54 to 60) and we don't want that.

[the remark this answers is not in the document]
tomaskrizek: It's fine, it's using `vagrantfile` variable. See https://github.com/freeipa/freeipa-pr-ci/pull/114/files#diff-7771668df07a2bb1307ced58ed0e0767R215 and https://github.com/freeipa/freeipa-pr-ci/pull/114/files#diff-7771668df07a2bb1307ced58ed0e0767R32, https://github.com/freeipa/freeipa-pr-ci/pull/114/files#diff-7771668df07a2bb1307ced58ed0e0767R205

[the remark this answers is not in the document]
felipevolpone: Does it make sense to have vagrantfile as a property? I don't see any external use of it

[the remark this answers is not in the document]
tomaskrizek: The important part is it enables proper use of inheritance. I see no reason to try and restrict the scope or visibility of `vagrantfile`.

github/prci.py
```diff
@@ -48,6 +48,7 @@ def __call__(self, depends_results={}):
         try:
             url = subprocess.check_output(cmd, shell=True)
         except subprocess.CalledProcessError as err:
+            logging.error(traceback.print_exc())
```
tomaskrizek: Is there a reason to use traceback module over `logging.error(err, exc_info=True)` ?

tasks/common.py
```diff
@@ -206,6 +208,26 @@ def logging_init_file_handler():
     LOG_FILE_HANDLER = fh
 
 
+def config_log_for_task(task_name, uuid):
+    """
+    task_name: name of the task in .freeipa-pr-ci.yaml
+    eg: fedora-26/caless
+    """
+    task_name = task_name.split('/')[1]
```
tomaskrizek: I'm not sure choosing the logger name dynamically based on task name is beneficial. There should always be just a single task in one directory, so it might be better to choose a static name like `job.log`. Then it's easier to find the log and we can avoid some weird collisions (`build.log` -- is it the output of mock or something else?).

tasks/tasks.py
```diff
@@ -55,6 +61,12 @@ def _before(self):
         # Create job dir
         try:
             os.makedirs(self.data_dir)
+            if self.task_name:
+                self.test_logger = config_log_for_task(self.task_name,
```
tomaskrizek: `self.test_logger` should be present in `__init__()`

github/prci_github/internals.py
```diff
@@ -503,20 +517,18 @@ def take_tasks(self):
         tasks = []
         for pull in PullRequests(self.repo):
             tasks_done_per_pr[pull.pull.number] = 0
-            for task in pull.tasks(self.tasks_config_path, self.job_cls):
+
+            pr_tasks = pull.tasks(self.tasks_config_path, self.job_cls)
+            logging.debug('Tasks in the PR: %s', [t.name for t in pr_tasks])
+            logging.debug('Tasks done PR: %s', tasks_done_per_pr)
```
tomaskrizek: I find logging this a bit odd. On each PR, it prints the number of tasks done for all previous PRs and always 0 for the current one, regardless of how many tasks are done. We should either drop this, log an accurate singel number for current PR or log it at the end of the loop for all PRs.

github/prci_github/internals.py
```diff
@@ -211,14 +216,23 @@ def take(self, runner_id):
...
     def execute(self, exc_handler=None):
         depends_results = {}
-        for dep in self.requires:
-            status = Status(self.repo, self.pull, dep)
-            depends_results[dep] = JobResult(
-                status.state, status.description, status.target_url)
+
+        # sometimes, when running runners in parallel, it may happen
+        # that we get an exception from OpenSLL.
+        @retry(OpenSSL.SSL.Error)
```
tomaskrizek: How does this work? It is reasonable to assume if it fails once, it's going to pass the next time around?

github/prci.py
```diff
@@ -252,17 +255,22 @@ def check_reboot(self, *args, **kwargs):
...
                     p.join()
+                    logging.info('Calling join() for process')
                     self.processes.remove(p)
 
         def execute_task(task):
             try:
                 task.execute()
-            except Exception:
+            except Exception as e:
+                logging.error(e)
```
tomaskrizek: We may or may not want `exc_info=True` here. I'm not sure.

github/prci.py
felipevolpone: I don't think so. Doing some tests and searching for it, it seems that there is no difference. We can shift to use `logging.error` in order to have just one way of doing it.

github/prci.py
felipevolpone: Actually, I think that here we could have: 
`logging.exception('Task %s failed to execute', task.name, exc_info=True)`

github/prci_github/internals.py
felipevolpone: The `retry` decorator will re-execute the method if the exception given as parameter is raised. 
In my many tests, it only happened to fail once, but as you said it could fail more times. We could change `retry` to receive the number of attempts that it should do. Something like: `@retry(OpenSSL.SSL.Error, attempts=2)`.

TBH I don't like the approach of having the `retry` decorator, however, given the deadlines, I think it's easier to go with it than change the code to use `threading`.

tasks/tasks.py
felipevolpone: `self.test_logger` is not in `__init__` because the `uuid` directory is created latter. So, if something happens between the class instantiation and the `_before` method be called, the log would point to a file that doesn't exist. 

As I prefer to change the code to do not create the log with the task name (following the comment above), I'll delete the code.

github/prci_github/internals.py
felipevolpone: Yeah, from my PVO is pretty much useless at this point. I forgot to remove that from my debugging.

tasks/tasks.py
tomaskrizek: I meant it should just be *defined*, even as `self.test_logger = None`. Just to make it clear this variable exists for the object and to satisfy pylint's desires (aka `attribute-defined-outside-init`).

felipevolpone:
As described in issue #54, this PR is being discontinued due to problems with running runners in parallel. PR #126  was created to address *only* the multiple topologies feature.

PR #115, opened 2017-10-30
Add missing backslash to README.md

Tiboris:
There was missing backslash in README.md


felipevolpone:
Good catch! Thanks @tiboris

Issue #105, opened 2017-10-17
Improve usability and sustainability of PR CI


tomaskrizek:
Make lives of users and developer of PR CI easier by addressing some of the minor issues.

## Benefits

- More usable, stable and robust interface to deal with
- Easier to use and maintain

## Tasks

**Ordered by priority**, not everything has to be done, some of the minor issues may be too difficult to implement and not worth it.

- [x] change `re-run` behavior to only re-run tasks in failed/error state #81, #83 
- [x] bring back logging of py.test's output #76, https://pagure.io/freeipa/issue/7186
- [x] do not require reboot during runner deployment #113 
- [x] update developer documentation (re-run) #93 
- [x] update developer documentation (ansible and vagrant) #97 
- [x] update README.md #107
- [x] unify replica numbering for logs and tests #74 
- [x] properly support hypervizor machines (`vmx` flag needed?) and test updated beaker file #89
- [x] improve help of `prci_test_control.py` #88
- [x] limit the size of systemd journal #53
- [x] move package list to a separate file #106 
- [x] improve cli for create-template-box #98
---
- [ ] runner disk cleanup #27 
- [ ] add a failover storage for logs #71 
- [ ] add informative messages to job status from py.test #17

felipevolpone:
@tomaskrizek don't you think that "Execute tasks in parallel to fully utilize runner" hasn't a high priority on this list too? IMHO it would be after "change re-run behaviour to only re-run..."

PR #117, opened 2017-10-30
Adding installed packages list to a separate file

### LINKED ISSUE #106: Move lists of installed packages from runner.log to separate file(s)
`runner.log.gz` is polluted with a list of installed packages for each machine (`machine/provision : get all packages`) and it's displayed twice (`machine/provision : display installed packages`).

This list can be useful, but it should be in a separate file(s), so the runner.log is readable.

felipevolpone:
With this change, the installed packages will be listed in the
/vagrant/installed_packages.log.gz files. After sending the
stdout to the file, we have a task that collects the files
in the shared directory. This is done via the shared directory
inside the vagrant VMs in `/vagrant`.
For more info about it, check `tasks/remote_storage.py::GzipLogFiles`

Fixes #106

### DIFF
--- a/ansible/roles/machine/provision/tasks/main.yml
+++ b/ansible/roles/machine/provision/tasks/main.yml
@@ -34,13 +34,13 @@
         - freeipa-*
         - python*-ipatests
 
-- name: get all packages
-  shell: rpm -qa
-  register: packages
+- name: create directory to save installed packages logs
+  file:
+    path: /vagrant/installed_packages/
+    state: directory
 
-- name: display installed packages
-  debug:
-    var: packages.stdout
+- name: get all packages
+  shell: rpm -qa > /vagrant/installed_packages/installed_packages_{{inventory_hostname}}.log
 
 - name: create hosts file from template
   template:


felipevolpone:
There is no rush to review this PR. It's a small contribution to "Improve usability and sustainability of PR CI" #105

felipevolpone:
The result (the logs) can be checked [here](https://github.com/felipevolpone/freeipa/pull/16)

ansible/roles/machine/provision/tasks/main.yml
```diff
@@ -35,12 +35,7 @@
         - python*-ipatests
 
 - name: get all packages
-  shell: rpm -qa
-  register: packages
-
-- name: display installed packages
-  debug:
-    var: packages.stdout
+  shell: rpm -qa > /vagrant/installed_packages.log
```
tomaskrizek: Previously, you could check the installed packages on all machines. They should be the same on all machines -- but we're losing the option to verify and check, because this overwrites the file for each VM. With the current setup, that's ok, since all packages are the same on all machines (same template, same playbook executed). If you're find with this behavior, ack.

tomaskrizek:
Do you have a PR run with this code so I don't have to deploy this myself? :)

felipevolpone:
You can check it here: https://github.com/felipevolpone/freeipa/pull/20

PR #118, opened 2017-11-02
Fixing the ipa-test-config.yaml in order to run external CA tests

felipevolpone:
To run the external CA tests, we have to provide the absolute path of the certs.


felipevolpone:
The result of this PR can be checked here: https://github.com/felipevolpone/freeipa/pull/19

PR #122, opened 2017-11-07
Force ipa-run-tests to use python 2

felipevolpone:
Since commit d39456a882efacbfefcc1f987ebf89ceeb761f61 in freeIPA, if freeIPA
is built using python3, ipa-run-tests will use python3 by default.
However, paramiko it's showing problems to collect logs from the hosts when using py3.

This is just a hotfix to have the logs back as soon as possible.

### DIFF
--- a/tasks/tasks.py
+++ b/tasks/tasks.py
@@ -225,7 +225,7 @@ class RunPytest(JobTask):
         self.execute_subtask(
             PopenTask(['vagrant', 'ssh', '-c', (
                 'IPATEST_YAML_CONFIG=/vagrant/ipa-test-config.yaml '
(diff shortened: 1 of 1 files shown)


felipevolpone:
Using ipa-run-tests-2 has solved the problem as we can see [here](https://fedorapeople.org/groups/freeipa/prci/jobs/ad9ad23e-c32c-11e7-8d06-00e08145c04a/).

However, I'm not sure if we want to get this PR merged. As I described in commit message, this is a hotfix to get the tests log back ASAP. I'm still looking for a proper solution (if there is one).

tomaskrizek:
We can always revert :) It's better to hotfix than to have missing logs.

PR #126, opened 2017-11-17
Enabling multiple topologies

### LINKED ISSUE #70: runner: support multiple topologies

felipevolpone:
In order to have more tests running in upstream PR CI
infrastructure, we need to support multiple topologies:
multiple master, replicas and clients. This PR address this.

To know the resources that a given task should have to
run it, the freeipa/freeipa/.freeipa-pr-ci.yml file will
declare how much RAM and CPUs are needed. Using this syntax:

```
topologies:
  master_1repl: &master_1repl
    name: master_1repl
    cpu: 2
    memory: 4098
    name: master_1repl_1client
    cpu: 6
    memory: 6144
```

After declaring the topologies, the tasks will be able
to use it with this syntax:

```
fedora-26/simple_replication:
    ...
    job:
      class: RunPytest
      args:
        ...
        topology: *master_1repl
```

Is IMPORTANT to say that, the name of the topology
must match with the name of a Vagrantfile inside the
templates/vagrantfiles dir. So, if a topology is added to
the .freeipa-pr-ci.yml file, a corresponding Vagrantfile
must be added to the freeipa-pr-ci project as well.

The Task class will read the config file and the Scheduler
will deal with it, to know if the runner is able to run
that task.

Fixes #70

### DIFF
--- a/github/prci_github/internals.py
+++ b/github/prci_github/internals.py
@@ -11,6 +11,7 @@ import pytz
 import requests
 import time
 import yaml
+import psutil
 
 import cachecontrol
 import redis
@@ -28,6 +29,10 @@ logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
 
 
 class TaskAlreadyTaken(Exception):
+    pass
+
+
+class InsufficientResources(Exception):
     pass
 
 
@@ -171,6 +176,7 @@ class Task(object):
 
         self.priority = conf['priority']
         self.requires = conf['requires']
+        self.resources = conf['job']['args'].get('topology', {})
         self.job = job_cls(conf['job'],
                            (self.repo.clone_url, self.refspec))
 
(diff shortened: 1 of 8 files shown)


felipevolpone:
The results can be checked [here](https://github.com/felipevolpone/freeipa/pull/26).
The task `fedora-26/caless_huge`  is not executed because there is no runner able to do it (due to the number of resources required).

github/prci_github/internals.py
```diff
@@ -369,6 +376,30 @@ def __init__(self, repo, tasks_config_path, job_cls, runner_id,
         self.tasks_config_path = tasks_config_path
         self.allowed_users = allowed_users
         self.runner_id = runner_id
+        self.total_cpus = psutil.cpu_count()
+        self.total_memory = psutil.virtual_memory().total / float(2 ** 20)
+        self.done = False
```
tomaskrizek: This variable isn't used anywhere.

PR #127, opened 2017-11-17
Limiting the systemd journal size

### LINKED ISSUE #53: runner: limit size of systemd journal
300MB should be enough for a couple of days. Large journal slows down some operations (systemctl status).

felipevolpone:
It's possible to change the value of 300M in the `defaults/main.yml` file.

Fixes #53

### DIFF
--- a/ansible/roles/runner/tasks/setup.yml
+++ b/ansible/roles/runner/tasks/setup.yml
@@ -67,3 +67,10 @@
     name: redis
     enabled: true
     state: started
+
+- name: configure size of systemd journal
+  lineinfile:
+    path: /etc/systemd/journald.conf
+    regexp: '^#?SystemMaxFileSize'
+    line: 'SystemMaxFileSize={{ limit_size_systemd_journal }}'
+    state: present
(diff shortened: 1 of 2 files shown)


ansible/roles/runner/tasks/setup.yml
```diff
@@ -67,3 +67,6 @@
     name: redis
     enabled: true
     state: started
+
+- name: configure size of systemd journal
+  shell: echo "SystemMaxFileSize={{ limit_size_systemd_journal }}" >> /etc/systemd/journald.conf
```
tomaskrizek: This adds the config line every time the playbook is executed. Please use the [lineinfile](http://docs.ansible.com/ansible/latest/lineinfile_module.html) module to avoid this.

ansible/roles/runner/tasks/setup.yml
```diff
@@ -67,3 +67,10 @@
     name: redis
     enabled: true
     state: started
+
+- name: configure size of systemd journal
+  lineinfile:
+    path: /etc/systemd/journald.conf
+    regexp: '^SystemMaxFileSize'
```
tomaskrizek: Since `SystemMaxFileSize` is commented out by default (and in the proper section), I'd use `regexp: '^#?SystemMaxFileSize'` to find the right place in the file.

felipevolpone: TBH I'm indifferent to it. Personally, I prefer to keep the config files with the original data and "override" the options. However, if you prefer this way I'm ok with it as well.

tomaskrizek: My concern is that this `SystemMaxFileSize` has to be in the `[Jounal]` section of the file. Since this is currently the only section, we can get away with just appending the value to the end of file, but this seems to be a bit more future-proof in case additional sections were added.

PR #129, opened 2017-11-20
do not require reboot during runner deployment

tomaskrizek:
Nested virt config tested with both kvm_intel and kvm_amd


tomaskrizek:
Also tested on system without nested virt support and it properly fails as expected.

felipevolpone:
Tested in scenarios:
* A working runner
* A new VM w/o nested virt. support
* A new VM w/ nested virt. support

Works great.

Issue #98, opened 2017-10-12
create-template-box CLI is confusing


felipevolpone:
The `create-template-box` script asks for inputs, however the CLI is confusing.
For example:
When `git branch (e.g. master):` is printed, it seems that if the user press enter (keeping it empty) the branch master will be used, but that is not true; and the validation (of empty field) is done later in the script.

So, we could:
* Use the example as default; or
* Check if it's empty and ask the user to input a valid information

PR #131, opened 2017-11-21
Adding default values to create-template-box CLI

felipevolpone:
Fixes: #98


ansible/create_template_box.yml
```diff
@@ -18,6 +18,7 @@
     - name: git_branch
       prompt: "git branch (e.g. master)"
```
tomaskrizek: I'd change the prompts to a more standard convetion indicating this is a default value: `git branch [master]`

felipevolpone: but that is how it works, if you execute the script, the prompt will be:
`git branch (e.g. master) [master]`

tomaskrizek: Ah, in that case, we could remove the `(e.g. master)`? :)

Issue #27, opened 2017-08-01
Runner disk cleanup


tomaskrizek:
Once a new template is generated and uploaded to vagrantcloud, there should be a (default) option to also revoke and older version from vagrant cloud and to delete it from all the runners via ansible. 

Two versions per template should be kept. One current stable and one updated one.

felipevolpone:
I think we could go with a script that runs after the upload script and ssh into the machines and removes the old boxes. What do you think? 
My concern is how to know every runner deployment we have.

tomaskrizek:
We need to solve 2 issues:

- revoking the old box -- we shouldn't need more than 2 active boxes (1 stable, 1 updated or deprecated). We should be very careful not to revoke a stable box. E.g. there is stable 9.0.1 and old 9.0.0. When we upload 9.0.2, we can delete 9.0.0. But then we find out we need more changes, so we create another update, 9.0.3. In this case, we *can't* remove the 9.0.1, but we should instead revoke 9.0.2.

- removing the old box from runners -- I don' think there's an ideal solution. I think we should just keep ansible inventory up to date and perhaps prompt the user to update it before confirming to proceed with the deletion of boxes.

PR #133, opened 2017-11-22
Script to cleanup vagrant boxes

felipevolpone:
Ansible script to remove vagrant boxes in hosts specified in runners inventory.

Fixes: #27

### DIFF
--- a/ansible/cleanup_runner_disk.yml
+++ b/ansible/cleanup_runner_disk.yml
@@ -1,0 +1,34 @@
+---
+- hosts: cleanup_runners
+  remote_user: root
+  name: cleanup runners disk
+  gather_facts: false
+
+  vars:
+    box_initial_name: "freeipa-VAGRANTSLASH"
+    default_provider: "libvirt"
+
+  vars_prompt:
+    - name: box_name
+      prompt: "Name of the box to be removed (eg: ci-master-f27)"
+      private: false
+
+    - name: box_version
+      prompt: "Version of the box to be removed (eg: 0.1.0)"
+      private: false
+
+    - name: user_confirmation
+      prompt: "Are you sure you want to remove this box? Answer with YES or NO"
+      default: "NO"
+      private: false
+
+  tasks:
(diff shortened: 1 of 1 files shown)


felipevolpone:
I've tried to keep it as simple as possible here.
I think that the best approach is we input which box version should be deleted, and not try to know which version should be revoked.

tomaskrizek:
I like keeping it simple -- managing the revocation manually seems to be a safer choice. However, this PR needs the following changes:

- two hard links exist for the image -- one in vagrant and the other in libvirt. Both need to be removed to actually free up space on disk.
- there needs to be the ability to remove individual versions of given boxes (e.g. freeipa/ci-master-f26, version 0.1.7)
- I think removing the boxes is safer by issuing the proper commands. It should work like this:
```bash
vagrant box remove --provider libvirt --box-version 0.1.7 -f freeipa/ci-master-f26
virsh vol-delete --pool default freeipa-VAGRANTSLASH-ci-master-f26_vagrant_box_image_0.1.7.img
```

felipevolpone:
sorry for the delay. changes addressed

ansible/cleanup_runner_disk.yml
```diff
@@ -0,0 +1,33 @@
...
+      default: "NO"
+      private: false
+
+  tasks:
+    - name: remove vagrant box
+      when: user_confirmation == "YES" and box_name != "" and box_version != ""
+      shell: vagrant box remove freeipa/{{ box_name }} --provider {{ default_provider }} --box-version {{ box_version }} 
+
+    - name: remove libvirt image
+      when: user_confirmation == "YES" and box_name != "" and box_version != ""
```
tomaskrizek: This would be a nice place for a [`block`](http://docs.ansible.com/ansible/latest/playbooks_blocks.html) to avoid copy-pasting the condition.

tomaskrizek:
Well, it works -- sort of... I have some improvements in other PR.

PR #136, opened 2017-11-28
CLI tool to open and close PRs

felipevolpone:
This PR:
* Adds a new CLI tool to be able to open and close PRs. Below, it's described how to use it.
* Adds an ansible script to configure a PR CI runner to automatically use the tool above to create PRs (using crontab)

<details>
  <summary>How to use the CLI tool</summary>

To open a PR:  
```
$ python3 github/open_close_pr.py --config config.yml open_pr
Running Open and Close PR Tool against felipevolpone/freeipa repo
Executing open_pr command
A new PR against felipevolpone/freeipa will be created with the title [Nightly Test PR] nt_20171128_104006
PR 55 created

$ python3 github/open_close_pr.py --config config.yml --prci_config ~/freeipa/.freeipa-pr-ci.yaml open_pr
```

To close a PR:  
```
$ python3 github/open_close_pr.py --config config.yml close_pr 55 --close_comment "Closing this PR"
Running Open and Close PR Tool against felipevolpone/freeipa repo
Executing close_pr command
PR 55 closed
```

The `config.yml` file must have:
```
repository:
    owner: felipevolpone
    name: freeipa
credentials:
    token: 123abctoken
```
</details>

<details>
<summary>Using the ansible script</summary>

```
$ ansible-playbook -i ansible/hosts/runners ansible/deploy_scheduler_runner.yml
You can provide a new .freeipa-pr-ci.yaml file to use it to open new PRs. Provide a path or just press enter: ~/projects/freeipa/.freeipa-pr-ci.yaml

PLAY [deploy scripts scheduler] ***********************************************************************************************************************

TASK [Gathering Facts] ********************************************************************************************************************************
ok: [1runner]

TASK [copy] *******************************************************************************************************************************************
changed: [1runner]

TASK [cron] *******************************************************************************************************************************************
changed: [1runner]

PLAY RECAP ********************************************************************************************************************************************
1runner                    : ok=3    changed=2    unreachable=0    failed=0
```
</details>

### DIFF
--- a/ansible/roles/automation/openclose_pr/tasks/main.yml
+++ b/ansible/roles/automation/openclose_pr/tasks/main.yml
@@ -1,0 +1,84 @@
+---
+- name: install packages dependencies
+  dnf:
+    name: "{{ item }}"
+    state: present
+  with_items:
+    - python3-github3py
+    - PyYAML
+
+- name: install pip dependencies
+  pip:
+    name: "{{ item }}"
+    executable: pip3
+  with_items:
+    - CacheControl
+    - gitpython
+
+- name: create base directory
+  file:
+    path: "{{ basedir }}"
+    state: directory
+    owner: root
+    group: root
+    mode: 0700
+
+- name: create configuration file
+  template:
+    src: config.yml
+    dest: "{{ basedir }}/config.yml"
+
+- name: copy openclose_pr.py script
+  copy:
+    src: "{{ playbook_dir }}/../github/open_close_pr.py"
+    dest: "{{ basedir }}"
+
+- name: copy custom freeipa-pr-ci.yaml file
+  copy:
+    src: "{{ prci_config }}"
+    dest: "{{ basedir }}/{{ identifier }}.yaml"
+
+- name: copy git ssh key
+  copy:
+    src: "{{ git_sshkey }}"
+    dest: "/root/.ssh/git_ssh_key"
+    mode: 0600
+    owner: root
+    group: root
+
+- name: configure ssh to use the right ssh key when using git
+  blockinfile:
+    path: /root/.ssh/config
+    create: yes
+    block: |
+          Host github-pr_tool
+            Hostname github.com
+            IdentityFile /root/.ssh/git_ssh_key
+
+- name: clone freeipa fork
+  shell: git clone git@github-pr_tool:{{ repo_owner }}/freeipa.git {{ basedir }}/freeipa
+
+# - name: clone freeipa fork
+#   git:
+#     repo: "git@openclose_tool.com:{{ repo_owner }}/freeipa.git"
+#     dest: "{{ basedir }}/freeipa"
+
+# this is done to avoid confusion between the original remote (freeipa/freeipa)
+- name: rename remote origin to mygithub
+  shell: "cd {{ basedir }}/freeipa; git remote rename origin mygithub"
+
+- cron:
+    name: "openclose_{{ identifier }}"
+    minute: "0"
+    hour: "23"
+    job: >
+           python3 {{ basedir }}/open_close_pr.py
+           --config {{ basedir }}/config.yml
+           --id {{ identifier }}
+           --repo_path {{ basedir }}/freeipa
+           --branch {{ branch }}
+           --prci_config {{  basedir }}/{{ identifier }}.yaml
+           --pr_against_upstream {{ pr_against_upstream | bool }}
+           open_pr
+    user: root
+    state: present
--- a/github/open_close_pr.py
+++ b/github/open_close_pr.py
@@ -1,0 +1,186 @@
+#!/usr/bin/python3
+
+import os
+import github3
+import argparse
+import logging
+import shutil
+import yaml
+from cachecontrol.adapter import CacheControlAdapter
+import git
+from git import Repo
+
+
+REF_FORMAT = 'refs/heads/'
+DEFAULT_COMMIT_MSG = 'automated commit'
+FREEIPA_PRCI_CONFIG_FILE = '.freeipa-pr-ci.yaml'
+REMOTE_REPO = 'https://github.com/freeipa/freeipa.git'
+UPSTREAM_REMOTE_REF = 'upstream'
+MYGITHUB_REMOTE_REF = 'mygithub'
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+consoleHandler = logging.StreamHandler()
+consoleHandler.setLevel(logging.INFO)
+logger.addHandler(consoleHandler)
+
+
+def load_yaml(yml_path):
+    try:
+        with open(yml_path) as yml_file:
+            return yaml.load(yml_file)
+    except IOError as exc:
+        raise argparse.ArgumentTypeError(
+            'Failed to open {}: {}'.format(yml_path, exc))
+    except yaml.YAMLError as exc:
+        raise argparse.ArgumentTypeError(
+            'Failed to parse YAML from {}: {}'.format(yml_path, exc))
+
+
+class AutomatedPR(object):
+
+    def __init__(self, github_token, repo):
+        github = github3.login(token=github_token)
+        github.session.mount('https://api.github.com', CacheControlAdapter())
+        self.repo = github.repository(repo['owner'], repo['name'])
+        self.upstream_repo = github.repository('freeipa', 'freeipa')
+
+    def commit_new_prci_config_file(self, new_prci_config, args):
+        """
+        Updates the .freeipa-pr-ci.yaml file with the content
+        of the --prci_config provided file.
+        """
+        prci_config_file = os.path.join(args.repo_path,
+                                        FREEIPA_PRCI_CONFIG_FILE)
+
+        shutil.copy(prci_config_file, new_prci_config)
+
+        repo = Repo(args.repo_path)
+        repo.git.add(FREEIPA_PRCI_CONFIG_FILE)
+        repo.git.commit('-m', DEFAULT_COMMIT_MSG)
+        repo.git.push(MYGITHUB_REMOTE_REF, args.id)
+
+    def close_older_pr(self, identifier):
+        refs = {r.ref:r for r in self.repo.refs()}
+        ref_uri = '{}{}'.format(REF_FORMAT, identifier)
+        ref = refs[ref_uri]
+        ref.delete()
+        logger.debug("Branch %s deleted", identifier)
+
+    def rebase_branch(self, base_branch, identifier, git_repo_path):
+        repo = Repo(git_repo_path)
+
+        try:
+            repo.git.remote('add', UPSTREAM_REMOTE_REF, REMOTE_REPO)
+        except git.exc.GitCommandError:
+            # the remote is already configured
+            pass
+
+        repo.git.fetch(UPSTREAM_REMOTE_REF)
+        repo.git.checkout('{}/{}'.format(UPSTREAM_REMOTE_REF, base_branch))
+
+        # creates new branch using the identifier as the name
+        repo.git.checkout('-b', identifier)
+        repo.git.push(MYGITHUB_REMOTE_REF, '-u', identifier)
+
+    def open_pr(self, args):
+        # before opening a new PR, we close the old one with the same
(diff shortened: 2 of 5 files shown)


github/open_close_pr.py
```diff
@@ -0,0 +1,132 @@
...
+
+        logger.info("A new PR against %s/%s will be created with "
+                    "the title %s", self.__repo.owner.login,
+                    self.__repo.source.name, pr_title)
+
+        head = '{}:{}'.format(self.__repo.owner.login, branch_name)
+        pr = self.__repo.create_pull(pr_title, 'master', head)
+        logger.info("PR %s created", pr.number)
+
+    def close_pr(self, args):
```
tomaskrizek: How do you propose we automate closing the PR when you need to specify the PR number? Would it be possible to detect it from branch name or title instead?

ansible/deploy_scheduler_runner.yml
```diff
@@ -0,0 +1,14 @@
+---
+- hosts: scheduler_runner
+  remote_user: root
+  name: deploy scripts scheduler
+  vars:
+    config_file: /root/.config/freeipa-pr-ci/config.yml
+  tasks:
+    - cron:
```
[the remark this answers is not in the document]
felipevolpone: I've assumed that we would not spawn a VM to only use it as a tool to open and close PRs. I've assumed that we would deploy this tool in some PR CI runner. That's why I'm not installing packages and assuming that `config.yml` is already there. I think that this is the way to go because it uses our resources in a better way.

github/open_close_pr.py
```diff
@@ -0,0 +1,132 @@
...
+DEFAULT_COMMIT_MSG = 'File for Nightly Tests'
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
+consoleHandler = logging.StreamHandler()
+consoleHandler.setLevel(logging.DEBUG)
+logger.addHandler(consoleHandler)
+
+
+class NightlyTests(object):
```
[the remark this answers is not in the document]
felipevolpone: good catch, thanks

github/open_close_pr.py
```diff
@@ -0,0 +1,132 @@
...
+        last_commit = next(self.__repo.commits())
+        branch_name = '{}_{}'.format(DEFAULT_BRANCH_NAME,
+                                     datetime.now().strftime("%Y%m%d_%H%M%S"))
+
+        new_branch_ref = '{}{}'.format(NEW_BRANCH_REF, branch_name)
+        self.__repo.create_ref(new_branch_ref, last_commit.sha)
+
+        return branch_name
+
+    def open_pr(self, args):
```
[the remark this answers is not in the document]
felipevolpone: I think in the same way. I've already done this in [this branch](https://github.com/felipevolpone/freeipa-pr-ci/commit/26654a4d632bd62b67906c80244050079044a7b1), however, I haven't included it in this PR yet. I'll update the PR to have it.

github/open_close_pr.py
felipevolpone: How I thought we could solve it: we'll need another tool that checks the status of opened PRs and produces the difference between the result of last Nightly Test PR and the current one. 

Probably, this tool will have to go over the opened PRs and "discover" which one is the "Nightly Tests PR", right? Then, it'll compare with the last result, add a comment with it and call the `close_pr` using the PR number.

What do you think?

felipevolpone:
We might have a problem that is when you run the ansible script twice for the same host. In the current state, a new cronjob entry will be created and this can lead to having two new PRs instead of one.

IMO, I could drop the use of the `cron` ansible command, and use `lineinfile`, however this way we would have to point the cron file: `/var/spool/cron/root` and add the entry manually.

However, I'm not an ansible expert, @tomaskrizek any thoughts on this?

tomaskrizek:
Just set a `name` for the [`cron`](http://docs.ansible.com/ansible/latest/cron_module.html) job and ansible will be able to properly detect it and won't create it again.

felipevolpone:
PR updated, please review the changes

github/open_close_pr.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def __init__(self, github_token, repo):
+        github = github3.login(token=github_token)
+        github.session.mount('https://api.github.com',
+                             CacheControlAdapter())
+
+        self.repo = github.repository(repo['owner'], repo['name'])
+
+    def create_branch(self, identifier):
+        last_commit = next(self.repo.commits())
```
tomaskrizek: Will this work if branch is something other than `master`? Or rather, if the latest commit is a in different branch than `args.branch`?

github/open_close_pr.py
```diff
@@ -0,0 +1,167 @@
...
+
+        for pull_request in self.repo.pull_requests():
+            if '[{}]'.format(pr_identifier) in pull_request.title:
+                ref_uri = '{}{}'.format(REF_FORMAT, pull_request.head.ref)
+                ref = refs[ref_uri]
+                ref.delete()
+                logger.debug("PR #%s closed and branch deleted",
+                             pull_request.number)
+
+    def rebase_branch(self, base_branch):
```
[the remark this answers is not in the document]
felipevolpone: yeah, I don't like either. However, I don't think it's possible to do the behavior above using the github3 lib. Probably, it would be necessary to use `Popen` and call the commands by ourselves. It's not a big deal anyway.

felipevolpone:
I have changed the way the rebase and the create branch process is done. the github3py library is quite limited (because GitHub API is quite limited). I didn't find a way to address the [changes you proposed](https://github.com/freeipa/freeipa-pr-ci/pull/136#discussion_r156078889), so I used `gitpython`. It's bad that this introduces another dependency, but from my POV is the way to go.

The downside of this implementation is that the user must provide GitHub API token and a ssh key. However, we could do the pull request process via command line as well, but is not so trivial as doing it via `github3py`.

github/open_close_pr.py
```diff
@@ -0,0 +1,185 @@
...
+        repo = Repo(args.repo_path)
+        repo.git.add(FREEIPA_PRCI_CONFIG_FILE)
+        repo.git.commit('-m', DEFAULT_COMMIT_MSG)
+        repo.git.push(MYGITHUB_REMOTE_REF, args.id)
+
+    def close_older_pr(self, pr_identifier):
+        # repo.git.push('-d', MYGITHUB_REMOTE_REF, pr_identifier)
+        # repo.git.branch('-D', identifier)
+
+        refs = {r.ref:r for r in self.repo.refs()}
```
felipevolpone: The code below could be replaced by the commented code.

felipevolpone:
Just to add more context to my proposal:
My idea is to we have a specific user (freeipa-automation or something like that; or maybe even use the freeipa-pr-ci user) that forks the freeipa repo and handles the automation processes.

ansible/roles/automation/openclose_pr/tasks/main.yml
```diff
@@ -0,0 +1,82 @@
...
+    dest: "/root/.ssh/git_ssh_key"
+    mode: 0600
+    owner: root
+    group: root
+
+- name: configure ssh to use the right ssh key when using git
+  blockinfile:
+    path: /root/.ssh/config
+    block: |
+          Host openclose_tool.com
```
tomaskrizek: Is there a reason to use a non-existent imaginary hostname? Why not `Host github.com`?

github/open_close_pr.py
```diff
@@ -0,0 +1,184 @@
...
+        """
+        prci_config_file = os.path.join(args.repo_path,
+                                        FREEIPA_PRCI_CONFIG_FILE)
+
+        content = ''
+        with open(new_prci_config, 'r') as f:
+            content = f.read()
+
+        with open(prci_config_file, 'w') as f:
+            f.write(content)
```
tomaskrizek: I think using [`shutil.copy`](https://docs.python.org/3/library/shutil.html#shutil.copy) is a cleaner solution.

github/open_close_pr.py
tomaskrizek: I believe it'd be more reliable to just delete the branch based on the identifier. I don't see any benefit in doing it by searching the PRs, just downsides (PR is missing or closed already for some reason; it depletes API requests; ...)

github/open_close_pr.py
```diff
@@ -0,0 +1,184 @@
...
+            repo.git.remote('add', UPSTREAM_REMOTE_REF, REMOTE_REPO)
+        except git.exc.GitCommandError:
+            # the remote is already configured
+            pass
+
+        repo.git.fetch(UPSTREAM_REMOTE_REF)
+        repo.git.checkout('{}/{}'.format(UPSTREAM_REMOTE_REF, base_branch))
+
+        # creates new branch using the identifier as the name
+        repo.git.checkout('-b', identifier)
```
tomaskrizek: This will fail if the branch wasn't deleted for some reason... Just something to keep in mind.

github/open_close_pr.py
```diff
@@ -0,0 +1,184 @@
...
+        self.commit_new_prci_config_file(args.prci_config, args)
+
+        pr_title = '[{}] Nightly PR'.format(args.id)
+
+        logger.debug("A new PR against %s/%s will be created with "
+                     "the title %s", self.repo.owner.login,
+                     self.repo.source.name, pr_title)
+
+        try:
+            pr = self.repo.create_pull(pr_title, args.branch, args.id)
```
tomaskrizek: There has to be an option to select which repo the PR should be opened against. Usually, a new branch will be created in the user's repo, but the nightly PR will be opened against official freeipa repo. This only allows me to open a PR against my own repo, which isn't very useful, since no PR CI machinery is running there.

ansible/roles/automation/openclose_pr/tasks/main.yml
felipevolpone: I have tried to keep the openclose tool as isolated as possible. So, if a machine configured with this tool is used by another tool for another purpose, the ssh key configuration would not become confusing (assuming that all tools should use the same key)

tomaskrizek: Ugh. Using imaginary domain name that might exist creates way more confusion.

If you want complete isolation on command-by-command basis, the do it properly with [`GIT_SSH_COMMAND`](https://stackoverflow.com/a/29754018/6081607).

If it's no necessary, you can use `Host github.com` to set defaults.

Or, just create a hostname like you did, but please do not use something that looks like actual domain name (e.g. `Host github-pr_tool` is fine).

felipevolpone:
Changes addressed

tomaskrizek:
@felipevolpone What about https://github.com/freeipa/freeipa-pr-ci/pull/136#discussion_r157733721 ?

ansible/roles/automation/openclose_pr/tasks/main.yml
felipevolpone: Ah ok, the problem you see is the hostname not with the use of it. Ok, cool. I'll change it.

felipevolpone:
@tomaskrizek I've missed that. I addressed the change and rebased the PR.

ansible/roles/automation/openclose_pr/tasks/main.yml
```diff
@@ -0,0 +1,84 @@
...
+    create: yes
+    block: |
+          Host github-pr_tool
+            Hostname github.com
+            IdentityFile /root/.ssh/git_ssh_key
+
+- name: clone freeipa fork
+  shell: git clone git@github-pr_tool:{{ repo_owner }}/freeipa.git {{ basedir }}/freeipa
+
+# - name: clone freeipa fork
```
tomaskrizek: Using `git` submodule seems like to proper way to clone the repo -- why is it commented out and a `shell` module is used instead?

felipevolpone: I was not able to make it work with a custom hostname and ssh key. 
In my tests in last week, it got stuck. So, I just used shell since it's just git clone.

felipevolpone:
@tomaskrizek hey, will you finish the review of this PR? otherwise, I'll merge it.

tomaskrizek:
@felipevolpone Sorry, I don't have the time (or resolve :D ) to do this review any more. It's up to you to decide how you want to proceed.

PR #138, opened 2017-11-30
create_template_box: configure pkg repos by distro version

tomaskrizek:
Enable per-version configuration of used package repositories. These
will be configured both in the template's system and in its mock build
envrionment.

---

This PR does a few things:
- It makes the user interface of creating a template a bit more user-friendly, because it no longer prompts the user twice for the same info.
- It makes it more user-friendly to generate a template for different fedora versions, supported are: `25`, `26`, `27` and `rawhide` and additional ones are easy to add in `ansible/vars/fedora/`.
- It enables per-distro version configuration of package repositories (e.g. I imagine we'll use `updates-testing` for branched, but yet unreleased fedoras)
- It properly configures the repos for both the template's system and in the mock build system that's used
- Enables creation of rawhide template ([ci-master-frawhide 0.0.2](https://app.vagrantup.com/freeipa/boxes/ci-master-frawhide/versions/0.0.2) was created using this PR)


ansible/roles/machine/setup/tasks/configuration.yml
```diff
@@ -4,8 +4,25 @@
...
+- name: enable selected repositories
+  shell: "dnf config-manager --set-enabled {{ item.key }}"
+  with_dict:
+    fedora: "{{ repo_fedora_enabled }}"
+    updates: "{{ repo_updates_enabled }}"
+    updates-testing: "{{ repo_updates_testing_enabled }}"
+  when: item.value
+
+- name: configure rawhide repository
+  shell: "dnf config-manager --set-{{ 'enabled' if repo_rawhide_enabled else 'disabled' }} rawhide"
```
felipevolpone: Isn't redundant check if `repo_rawhide_enabled` is true? IIUC in all situations where `fedora_version` is `"rawhide"`,`repo_rawhide_enabled` is also true. There is some situation where `fedora_version` is `"rawhide"` and the repo should not be enabled?

ansible/vars/fedora/25.yml
```diff
@@ -0,0 +1,8 @@
+---
```
felipevolpone: Could we have `repo_rawhide_enabled: 0` here as well? It would be nice if all `fedora/<version>.yml` files would have all options, but each one set with the right value.

ansible/vars/branches/ipa-4-5.yml
```diff
@@ -0,0 +1,2 @@
+---
```
felipevolpone: Could we change the name of the directory from `branches` to `ipa-branches`?

ansible/roles/builder/prepare/tasks/setup_mock.yml
```diff
@@ -1,38 +1,9 @@
 ---
```
felipevolpone: Could we join this file with the `install_packages.yml` file? Both are small and could be together inside `main.yml`. This way we remove another "indirection layer".

felipevolpone:
@tomaskrizek Another two minor suggestions:

* Could you, please, add in `doc/README.md` instructions on what should be done if we would like to support another Fedora version?

* Shouldn't this code in `roles/machine/provision/tasks/main.yml` file be removed?
```
roles/machine/provision/tasks/main.yml
  rescue:
    # Distro repos are turned off in default template
    # If a new package from fedora/updates is required, this will fix it
    - dnf:
        name: "{{ item }}"
        state: latest
        enablerepo: fedora,updates
      with_items:
        - freeipa-*
        - python*-ipatests
```

ansible/vars/fedora/25.yml
tomaskrizek: What would be the point of adding it to non-rawhide template? It'd just be an unused variable. You could set it to `0`, `1`, or `"This is a value"` and it wouldn't have any effect.

ansible/roles/machine/setup/tasks/configuration.yml
tomaskrizek: 1. I wanted to specifically enable it instead of relying on the defaults.
2. I wanted to have an option to disable it, if we want to do so in the future. (for whatever reason, we just one day decide -- hey, let's just freeze rawhide packages for a week)

If you don't agree, I'm not against removing it, but I think it limits our options for no reason.

ansible/vars/fedora/25.yml
tomaskrizek: Given what I said, I'm removing the `fedora`, `updates` and `updates-testing` from the rawhide variables.. It doesn't make any sense to have them there either.

tomaskrizek:
I've updated the PR, please review the changes.

ansible/vars/fedora/25.yml
felipevolpone: The point would be to explicitly show all options that should be configured in each case. The variable not used is harmless, compared to the "documentation POV".

I'll not NACK the PR because of that, though.

ansible/roles/box/prepare/templates/Vagrantfile.j2
```diff
@@ -6,7 +6,7 @@ Vagrant.configure(2) do |config|
 
     config.vm.provider "libvirt" do |domain, override|
         override.vm.box_url = "{{ base_box_url }}"
-        override.vm.box = "{{ base_box_name }}"
+        override.vm.box = "f{{ fedora_version }}"
```
felipevolpone: Don't we have a typo here? Shouldn't be `"{{ fedora_version }}"` (without the `f')?

ansible/vars/fedora/25.yml
tomaskrizek: My point is, for rawhide, these variables do not need to be and should not be configure. Same thing for branched, except with a different repo. Rawhide is a special case of Fedora, so it's natural to treat it a bit differently.

felipevolpone:
The [rawhide image](http://ftp.fi.muni.cz/pub/linux/fedora/linux/development/rawhide/CloudImages/x86_64/images/Fedora-Cloud-Base-Vagrant-Rawhide-20171129.n.0.x86_64.vagrant-libvirt.box) is returning 404.

felipevolpone:
Testing with fedora 25 released the error:
```
$ ./create-template-box
Using /etc/ansible/ansible.cfg as config file
SUDO password:
fedora [27]: 25
git branch [master]:
```

<details>
<summary>end of the log</summary>

```
TASK [builder/build : create srpm] ***************************************************************************************************************************************
...

cmd: mock --buildsrpm --spec /root/rpmbuild/SPECS/freeipa.spec --sources /root/rpmbuild/SOURCES/freeipa-4.6.90test.tar.gz --result /root/rpmbuild/SRPMS/

start: 2017-12-04 20:53:02.697243

end: 2017-12-04 20:53:03.455153

delta: 0:00:00.757910

stderr: INFO: mock.py version 1.4.7 starting (python version = 3.5.2)...
Start: init plugins
INFO: tmpfs initialized
INFO: selinux disabled
Finish: init plugins
Start: run
INFO: Start(/root/rpmbuild/SPECS/freeipa.spec)  Config(fedora-25-x86_64)
Start: clean chroot
Finish: clean chroot
Start: chroot init
INFO: mounting tmpfs at /var/lib/mock/fedora-25-x86_64/root.
INFO: calling preinit hooks
INFO: enabled root cache
INFO: enabled dnf cache
Start: cleaning dnf metadata
Finish: cleaning dnf metadata
INFO: enabled HW Info plugin
Mock Version: 1.4.7
INFO: Mock Version: 1.4.7
Start: dnf install
ERROR: Exception(/root/rpmbuild/SPECS/freeipa.spec) Config(fedora-25-x86_64) 0 minutes 0 seconds
INFO: Results and/or logs in: /root/rpmbuild/SRPMS/
INFO: Cleaning up build root ('cleanup_on_failure=True')
Start: clean chroot
INFO: unmounting tmpfs.
Finish: clean chroot
ERROR: Command failed:
 # /usr/bin/dnf --installroot /var/lib/mock/fedora-25-x86_64/root/ --releasever 25 --disableplugin=local --setopt=deltarpm=False install @buildsys-build
Repository 'freeipa_copr': Error parsing config: Error parsing "enabled = 1  # enabled-switch:freeipa_copr": invalid boolean value
Repository 'fedora': Error parsing config: Error parsing "enabled = 1  # enabled-switch:fedora": invalid boolean value
Repository 'updates': Error parsing config: Error parsing "enabled = 1  # enabled-switch:updates": invalid boolean value
Repository 'updates-testing': Error parsing config: Error parsing "enabled = 0  # enabled-switch:updates-testing": invalid boolean value
Error: There are no enabled repos.
        to retry, use: --limit @/home/fbarreto/projects/freeipa-pr-ci/ansible/create_template_box.retry

PLAY RECAP ***************************************************************************************************************************************************************
192.168.121.206            : ok=21   changed=16   unreachable=0    failed=1
localhost                  : ok=11   changed=7    unreachable=0    failed=0
```
</details>

tomaskrizek:
Latest available rawhide nightly box is now properly detected and downloaded when creating rawhide template.

I've also fixed the mock config files. However, we still can't create `ci-f25-master`, because we no longer have up-to-date dependencies in copr. I think we can just update the supported (in working state) templates and forget about the rest.

Issue #109, opened 2017-10-18
Run tests on rawhide fedora


felipevolpone:
Often happens that we have IPA with some errors when running in a fedora rawhide. We could run nightly tests using rawhide fedora and compare the results with the last run. 

## Benefits
* Figure out problems before releases
* Discover when a new version of some dependency breaks IPA

## Tasks
- [ ] Add an option to disable COPR during template creation (PR #138)
- [ ] Add option to always update pkgs (both in mock and system) during provisioning (PR #139)
- [ ] Automate the creation of PRs at night (PR #136)
- [x] Create a fedora rawhide template ([ci-master-frawhide](https://app.vagrantup.com/freeipa/boxes/ci-master-frawhide))
- [ ] Create a tool to compare results and pkgs with previous test run and generate diff in the PR comment (see https://github.com/tomaskrizek/freeipa/pull/35#issuecomment-327547332 for example)
- [ ] Create a tool to open and close PRs (PR #136)

## Depends on
- ~Use PR CI infrastructure to run nightly tests, #108~

felipevolpone:
PoC:
* https://github.com/tomaskrizek/freeipa/pull/35
* https://github.com/tomaskrizek/freeipa-pr-ci/tree/rawhide-nightly

tomaskrizek:
We don't actually need to create a nigthly rawhide template. It's enough to do this once per release (already done for [ci-master-f28](https://app.vagrantup.com/freeipa/boxes/ci-master-f28)). Then, every night, we just update the packages. It's just important not to use our COPR and use clean distro packages.

The most time consuming part would probably be to automate generating diffs of updated packages and broken/fixed tests since last test. I imagine the result could look like https://github.com/tomaskrizek/freeipa/pull/35#issuecomment-327547332

tomaskrizek:
Rawhide testing would allow us to discover issues with released fedora packages early, so let's do this sooner rather than later.

rcritten:
This is a good idea but in my experience in developing in rawhide for the past few months rawhide is very prone to work-stopping package issues. Yes you catch things earlier but this can't be a blocker for accepting patches because it may have nothing to do with the individual patch (or it might).

IMHO rawhide should be only rawhide and not rely on copr repos to make things work.

tomaskrizek:
[...]

PR #139, opened 2017-12-01
VM provisioning: add option to update packages

tomaskrizek:
Enable a package update during VM provisioning in RunPytest job.

Related #109

### DIFF
--- a/tasks/tasks.py
+++ b/tasks/tasks.py
@@ -189,10 +189,12 @@ class RunPytest(JobTask):
     action_name = 'run_pytest'
 
     def __init__(self, template, build_url, test_suite, topology=None,
-                 timeout=constants.RUN_PYTEST_TIMEOUT, **kwargs):
+                 timeout=constants.RUN_PYTEST_TIMEOUT, update_packages=False,
+                 **kwargs):
         super(RunPytest, self).__init__(template, timeout=timeout, **kwargs)
         self.build_url = build_url + '/'
         self.test_suite = test_suite
+        self.update_packages = update_packages
 
         if not topology:
             topology = {'name': constants.DEFAULT_TOPOLOGY}
@@ -214,7 +216,8 @@ class RunPytest(JobTask):
                     action_name=self.action_name),
                 os.path.join(self.data_dir, 'vars.yml'),
                 dict(repofile_url=urllib.parse.urljoin(
-                        self.build_url, 'rpms/freeipa-prci.repo')))
+                        self.build_url, 'rpms/freeipa-prci.repo'),
+                     update_packages=self.update_packages))
         except (OSError, IOError) as exc:
             msg = "Failed to prepare test config files"
             logging.debug(exc, exc_info=True)
--- a/ansible/roles/machine/provision/tasks/main.yml
+++ b/ansible/roles/machine/provision/tasks/main.yml
@@ -15,6 +15,12 @@
     dest: /etc/yum.repos.d/
     url: "{{ repofile_url }}"
 
+- name: update packages
+  dnf:
+    name: '*'
+    state: latest
+  when: update_packages is defined and update_packages
+
 - name: install freeipa packages
   block:
     - dnf:
--- a/templates/run_pytest.vars.yml
+++ b/templates/run_pytest.vars.yml
@@ -1,2 +1,3 @@
 ---
 repofile_url: {{ repofile_url }}
+update_packages: {{ update_packages }}


tasks/tasks.py
```diff
@@ -189,10 +189,12 @@ class RunPytest(JobTask):
     action_name = 'run_pytest'
 
     def __init__(self, template, build_url, test_suite, topology=None,
-                 timeout=constants.RUN_PYTEST_TIMEOUT, **kwargs):
+                 timeout=constants.RUN_PYTEST_TIMEOUT, update_packages=False,
+                 **kwargs):
         super(RunPytest, self).__init__(template, timeout=timeout, **kwargs)
         self.build_url = build_url + '/'
         self.test_suite = test_suite
+        self.update_packages = update_packages
```
felipevolpone: Can we change this to `__update_packages` since it's not used out of the class?

tasks/tasks.py
```diff
@@ -189,10 +189,12 @@ class RunPytest(JobTask):
     action_name = 'run_pytest'
 
     def __init__(self, template, build_url, test_suite, topology=None,
-                 timeout=constants.RUN_PYTEST_TIMEOUT, **kwargs):
+                 timeout=constants.RUN_PYTEST_TIMEOUT, update_packages=False,
```
felipevolpone: Shouldn't be `False` instead of `false`?

felipevolpone:
@tomaskrizek some minor comments inline. Also, could you, please, update the commit message to explain why this change is necessary?

tasks/tasks.py
felipevolpone: It's ok. I haven't refreshed the page and commented on an outdate code.

tasks/tasks.py
felipevolpone: I don't see your argument here. Could you provide it in some constructively way? From my POV, there is no need to provide details of the implementation, if the attribute is not part of the API.

stlaz: What's the problem with it being "public" and you being able to change its value in the lifetime of the object?
Your `build_url` and `test_suite` variable are "public", why suddenly change the new variable to be private?

tomaskrizek:
You can see it updates the packages in this job: https://fedorapeople.org/groups/freeipa/prci/jobs/16a786a0-d930-11e7-aab2-6cf04961d829/runner.log.gz

tasks/tasks.py
felipevolpone: That's the point, IMO `build_url` and `test_suite` should be private as well. If objects don't have their state changed, the code is easier to debug, understand and maintain.

tomaskrizek: Is there a reason to hide it? I think not. On the contrary, I could imagine we might want to use the variable in some derived class. I don't think it's Pythonic to hide everything you can and it actually reduces readability.

Even if there was a reason to hide/protect it, I'd go with a single underscore, to just indicate this is something you shouldn't mess with, but to not invoke name mangling. These are used for example for `_run()` and similar methods, because you should use the object as callable instead of invoking these methods manually.

felipevolpone: > Is there a reason to hide it? I think not.

IMHO, you should have a reason to expose not the opposite. The API should be the simplest possible from the user POV. So, if there are too many attributes/methods/whatever, it's an indicator that your code is not easy to use. Anyway, we can go with one underscore then.

davidkupka: IMO the default visibility (public vs _hidden vs __protected) may differ project by project. So, first choose you default and then use it consistently in the project.
I (and this is subjective)  don't like __protecting attributes unless there's reason for it but don't have special feelings for public and _hidden attributes.

tomaskrizek: There is no precedent in using `_` or `__` in attribute names in the `tasks` module.

@felipevolpone If you'd like to change this, please move this discussion and change to a separate PR, which will address **all** the same cases, to be consistent. I found at least:
- `Task.tasks`
- `Task.exc`
- `FallibleTask.raise_on_err`
- `PopenTask.env`
- `PopenTask.process`
- `JobTask.template_name`
- `JobTask.template_version`
- `JobTask.remote_url`
- `Build.git_respec`
- `Build.git_version`
- `Build.git_repo`
- `RunPytest.build_url`
- `RunPytest.test_suite`
- `VagrantBoxDownload.box`
- `VagrantBox.name`
- `VagrantBox.version`
- `VagrantBox.provider`
- `AnsibleFixKeysPermissions.directory`
- `AnsiblePlaybook.extra_vars`
- `AnsiblePlaybook.playbook`
- `AnsiblePlaybook.verbosity`

Issue #141, opened 2017-12-04
attemp to restart the prci service if it fails


tomaskrizek:
In case `api.github.com` is temporarily inaccessible, if runners happen to attempt to start the service at that time, it will fail (complete traceback attached below). Incidentally, this also disables automatic reboots, which can make the machine go stale and become inaccessible through ssh.

We can either use systemd to restart the service if it fails, or handle the exceptions during `HTTPSConnectionPool` initialization. I'm in favor of using the systemd service restart, since it can also cover other failures.

Traceback:
```
Traceback (most recent call last):
   File "/usr/lib/python3.6/site-packages/requests/packages/urllib3/connection.py", line 141, in _new_conn
     (self.host, self.port), self.timeout, **extra_kw)
   File "/usr/lib/python3.6/site-packages/requests/packages/urllib3/util/connection.py", line 60, in create_connection
     for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
   File "/usr/lib64/python3.6/socket.py", line 745, in getaddrinfo
     for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
 socket.gaierror: [Errno -2] Name or service not known
 During handling of the above exception, another exception occurred:
 Traceback (most recent call last):
   File "/usr/lib/python3.6/site-packages/requests/packages/urllib3/connectionpool.py", line 600, in urlopen
     chunked=chunked)
   File "/usr/lib/python3.6/site-packages/requests/packages/urllib3/connectionpool.py", line 345, in _make_request
     self._validate_conn(conn)
   File "/usr/lib/python3.6/site-packages/requests/packages/urllib3/connectionpool.py", line 844, in _validate_conn
     conn.connect()
   File "/usr/lib/python3.6/site-packages/requests/packages/urllib3/connection.py", line 284, in connect
     conn = self._new_conn()
   File "/usr/lib/python3.6/site-packages/requests/packages/urllib3/connection.py", line 150, in _new_conn
     self, "Failed to establish a new connection: %s" % e)
 requests.packages.urllib3.exceptions.NewConnectionError: <requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x7f16f7a08940>: Failed to establish a new connection: [Errno -2] Name or service not known
 During handling of the above exception, another exception occurred:
 Traceback (most recent call last):
   File "/usr/lib/python3.6/site-packages/requests/adapters.py", line 423, in send
     timeout=timeout
[...]

PR #142, opened 2017-12-04
prci.service: auto-restart service on failure

tomaskrizek:
Configure prci.service to be restarted after 10 minutes if it fails.
Rate-limiting of startup is disabled, so systemd will attempt to restart
the service indefinitely as long as it fails.

Fixes: #141

### DIFF
--- a/ansible/roles/runner/files/prci.service
+++ b/ansible/roles/runner/files/prci.service
@@ -1,6 +1,7 @@
 [Unit]
 Description=FreeIPA PR CI
 After=syslog.target network-online.target libvirtd.service
+StartLimitIntervalSec=0
 
 [Service]
 Type=simple
@@ -10,6 +11,8 @@ WorkingDirectory=/root/freeipa-pr-ci
 ExecStart=/bin/bash -c '/root/freeipa-pr-ci/github/prci.py "$(hostname -s)" --config /root/.config/freeipa-pr-ci/config.yml'
 StandardOutput=syslog
 StandardError=syslog
+Restart=on-failure
+RestartSec=10m
 
 [Install]
 WantedBy=multi-user.target


ansible/roles/runner/files/prci.service
```diff
@@ -10,6 +10,9 @@ WorkingDirectory=/root/freeipa-pr-ci
 ExecStart=/bin/bash -c '/root/freeipa-pr-ci/github/prci.py "$(hostname -s)" --config /root/.config/freeipa-pr-ci/config.yml'
 StandardOutput=syslog
 StandardError=syslog
+Restart=on-failure
+RestartSec=10m
```
felipevolpone: There is some reason behind the magical `10m` value?

tomaskrizek: Not really, 10 minutes seemed like an appropriate time to reattempt restart. In case the service is unreachable, there's no need to try more often than that. In cases there's some other error which depletes some API requests, it seems like limiting it to 6 times per hour is reasonable so it doesn't affect other runners.

felipevolpone: Nice, I was just wondering if I was missing the why.

ansible/roles/runner/files/prci.service
```diff
@@ -10,6 +10,9 @@ WorkingDirectory=/root/freeipa-pr-ci
...
+StartLimitIntervalSec=0
```
felipevolpone:
```
Dec 05 10:55:29 .... systemd[1]: [/etc/systemd/system/prci.service:15] Unknown lvalue 'StartLimitIntervalSec' in section 'Service'
```
`StartLimitIntervalSec` belongs to the `[Unit]` section.

tomaskrizek: Ah, that's right.. Thanks for the catch.

PR #152, opened 2018-01-24
Changing WebUI config to also run tests that depends on KRA

felipevolpone:


felipevolpone:
Since this is a very trivial PR, I'm merging it.

PR #153, opened 2018-01-25
Removing use of custom COPR repo

felipevolpone:
Instead of using custom COPR repos, get the packages from pip, since we have the required version there.


tomaskrizek:
LGTM. `CacheControl>=0.12.3` should also work. It was just the older version which had some caching issues.

PR #154, opened 2018-01-30
Creating more templates to run tests with 2 and 3 replicas

felipevolpone:


felipevolpone:
Merging it since it's a trivial change

PR #155, opened 2018-02-06
Changing to install a newer Vagrant version

felipevolpone:
As said here [1], a newer version of Vagrant is needed. Otherwise, the process to add a box will fail with a 404, because the "server/repo URL" has been moved.

* Installing new Vagrant version
Instead of relying on fedora packages, this changes the code to get the rpm directly from vagrantcloud source. I prefer to maintain a simple URL that can be easily changed when it's needed.

* Adding the necessary packages to the new version
As can be checked here: https://github.com/vagrant-libvirt/vagrant-libvirt

* This commit also specify the provider (libvirt in our case) to use in
vagrant up, as it's needed by the new vagrant version.

[1] https://groups.google.com/forum/#!msg/vagrant-up/H8C68UTkosU/qz4YUmAgBAAJ

### DIFF
--- a/templates/vagrantfiles/Vagrantfile.build
+++ b/templates/vagrantfiles/Vagrantfile.build
@@ -37,3 +37,5 @@ Vagrant.configure(2) do |config|
     end
 
 end
+
+Vagrant::DEFAULT_SERVER_URL.replace('https://vagrantcloud.com')
--- a/templates/vagrantfiles/Vagrantfile.ipaserver
+++ b/templates/vagrantfiles/Vagrantfile.ipaserver
@@ -44,3 +44,5 @@ Vagrant.configure(2) do |config|
         end
     end
 end
+
+Vagrant::DEFAULT_SERVER_URL.replace('https://vagrantcloud.com')
--- a/templates/vagrantfiles/Vagrantfile.master_1repl
+++ b/templates/vagrantfiles/Vagrantfile.master_1repl
@@ -60,3 +60,5 @@ Vagrant.configure(2) do |config|
     end
 
 end
+
+Vagrant::DEFAULT_SERVER_URL.replace('https://vagrantcloud.com')
--- a/templates/vagrantfiles/Vagrantfile.master_1repl_1client
+++ b/templates/vagrantfiles/Vagrantfile.master_1repl_1client
@@ -58,3 +58,5 @@ Vagrant.configure(2) do |config|
     end
 
 end
+
+Vagrant::DEFAULT_SERVER_URL.replace('https://vagrantcloud.com')
--- a/templates/vagrantfiles/Vagrantfile.master_2repl_1client
+++ b/templates/vagrantfiles/Vagrantfile.master_2repl_1client
@@ -61,3 +61,5 @@ Vagrant.configure(2) do |config|
     end
 
 end
+
+Vagrant::DEFAULT_SERVER_URL.replace('https://vagrantcloud.com')
--- a/templates/vagrantfiles/Vagrantfile.master_3repl_1client
+++ b/templates/vagrantfiles/Vagrantfile.master_3repl_1client
@@ -64,3 +64,5 @@ Vagrant.configure(2) do |config|
     end
 
 end
+
+Vagrant::DEFAULT_SERVER_URL.replace('https://vagrantcloud.com')


rcritten:
Is this because Fedora is lagging behind the vagrant packaged version?

felipevolpone:
Kind of. Would be nice to use vagrant 2.x anyway, but yes, the change to get the package from the website is because fedora 26 provides vagrant 1.9.1.

pvoborni:
We'll also need the upstream version for AD vagrant box: https://app.vagrantup.com/peru/boxes/windows-server-2016-standard-x64-eval

stlaz:
Fedora 27 provides vagrant 2.0.1. Just upgrade your infrastructure.

felipevolpone:
We can't update the infrastructure to f27 right now. We got tons of errors during `vagrant up`due to some problem with dnsmasq IIRC. I don't have the exception/error log at the hand right now, but you can try it by yourself creating a runner with the f27 template provided by openstack.

stlaz:
Why don't you track the issues? They sound kind of severe. Solving them could move this project forward and uncover possible bugs instead of trying to add a workaround using distro-unsupported package.

If this blocks the builds, you can try the other solution proposed in the discussion mentioned, and that is setting the "VAGRANT_SERVER_URL" environment variable, while Fedora 27 fixes are being worked on.

Also, please, stop doing multiple unrelated changes in one commit in case you need to either revert it or use `git bisect` to uncover bugs.

felipevolpone:
Since I'm running the project on my own, I prefer to choose stability instead of the risk of PR CI blowing up anytime.

So, we will keep in f26; and vagrant 2.x is needed for AD tests, as @pvoborni already said.

> Also, please, stop doing multiple unrelated changes in one commit in case you need to either revert it or use git bisect to uncover bugs.

They are not unrelated. All changes are part of the migration to use vagrant 2.x and should be applied atomically.

ansible/roles/runner/tasks/setup.yml
```diff
@@ -46,6 +47,9 @@
   notify:
     - restart_nfs
 
+- name: installing libvirt plugin for vagrant
+  shell: vagrant plugin install vagrant-libvirt
```
tomaskrizek: If you want to install the vagrant-libvirt plugin in this way, it'd probably be best if you'd remove the `vagrant-libvirt` from package installation above (line 22) to avoid any confusion.

stlaz:
Ah, sorry, I noticed those unrelated changes in one commit several times before and the structure of this commit got me confused. Take it as a general advice, then.

PRCI may blow up anytime on F26 and F27 just as well, F27 has been there for enough time to be stable now. My note was to simplify future maintenance for you or anyone who'll be working with or in this project by choosing platform supported packages and it is of course up to you to decide how are you going to deal with my comments.

We did not need AD boxes up until now and I would prioritize fixing current issues (read as those on Fedora 27) before adding new features.

All of this is of course my opinions and you may agree or disagree. Since I don't mean to maintain this project myself, it's your choice.

felipevolpone:
> PRCI may blow up anytime on F26 and F27 just as well, F27 has been there for enough time to be stable now. My note was to simplify future maintenance for you or anyone who'll be working with or in this project by choosing platform supported packages and it is of course up to you to decide how are you going to deal with my comments.

Don't get me wrong, I agree with you. It would be much easier just update to f27, but f27 templates in openstack are not stable. I didn't choose the approach to use `VAGRANT_SERVER_URL` because once we start running in f27 (and using vagrant 2.x), we'll need to add the `libvirt-devel` package, add the installation of the vagrant plugin and add the `--provider` anyway. So I'm anticipating the changes we will need.

It seems easier to me, in the future, create a PR that changes back the URL (to get vagrant 2.x) to just `vagrant`.

> All of this is of course my opinions and you may agree or disagree. Since I don't mean to maintain this project myself, it's your choice.

I didn't say it to be rude, but from the POV from someone who sometimes has to put out the fire by itself. Your comments and feedbacks are very welcome :)

felipevolpone:
Just forget everything that I said. Let's do it in the easiest way :) 

JFTR the problem with f27 is [this one](https://fedorapeople.org/groups/freeipa/prci/jobs/0da2a7a0-0c30-11e8-93b6-fa163ee18643/runner.log.gz):
```
Error while activating network: Call to virNetworkCreate failed: failed to 
run '/usr/sbin/dnsmasq --version': : No such file or directory.
```

More details:
```
[root@prci-os-temp4 ~]# whereis dnsmasq
dnsmasq: /usr/sbin/dnsmasq /etc/dnsmasq.d /etc/dnsmasq.conf /usr/share/dnsmasq /usr/share/man/man8/dnsmasq.8.gz
[root@prci-os-temp4 ~]# rpm -qa | grep dnsmasq
dnsmasq-2.78-2.fc27.x86_64
[root@prci-os-temp4 ~]# dnsmasq --versiopn
dnsmasq: relocation error: dnsmasq: symbol nettle_get_hashes, version NETTLE_6 not defined in file libnettle.so.6 with link time reference
[root@prci-os-temp4 ~]# dnsmasq --version
dnsmasq: relocation error: dnsmasq: symbol nettle_get_hashes, version NETTLE_6 not defined in file libnettle.so.6 with link time reference
```

I'm still investigating it, any help is very welcome.

stlaz:
This may be a bug in dnsmasq package requirements. The combination of
```
nettle-3.4-1.fc27.x86_64
dnsmasq-2.78-2.fc27.x86_64
```
works fine for me.

PR #156, opened 2018-02-06
Increase the time between requests

felipevolpone:
As commit a811f1cc explains, we need to slow down the API calls to avoid
hitting the limit of requests per minute. To have more than ~20 runners, it's necessary to increase this time. Increasing it to 1.5s would allow us to have around 40 runners.

Important to say that the limit is hit even when runners are using the cached requests, as it's well explained in [1].

[1] https://github.com/freeipa/freeipa-pr-ci/tree/master/doc#abuse-limit

### DIFF
--- a/github/prci_github/adapter.py
+++ b/github/prci_github/adapter.py
@@ -36,7 +36,7 @@ class GitHubAdapter(CacheControlAdapter):
             logger.debug('%s: try %d', self.__class__.__name__, try_counter)
 
             # Rate-limit requests to avoid hitting GitHub API abuse limit
-            time.sleep(0.5)
+            time.sleep(1.5)
 
             try:
                 response = super(GitHubAdapter, self).send(


tomaskrizek:
This will probably significantly increase the time until runners pick up a new task, since they iterate over the whole PR queue until they do.

This limit is mainly an issue when the runners are idle and iterating over the PR queue to find new tasks. I think you'd be able to lower this limit with the following approach:

- start with a low time delay (or none)
- if you get limit exceeded error, increase the delay
- if the limit isn't exceeded for a period of time, lower the delay

I think this adaptive approach would provide best results with the current design. Credit for this idea goes to @davidkupka :)

felipevolpone:
> This will probably significantly increase the time until runners pick up a new task, since they iterate over the whole PR queue until they do.

For sure. I see this PR as a hotfix to the problem. The ideal would be a rewrite the architecture of PR CI, as we already discussed some day in the past.

About the adaptative approach I'm not sure if it would make a real difference, because the second case would happen very very soon and the third could not happen at all. Right now, with ~27 runners, we hit the 1500 req/min limit in minutes (even when providing a dump of the cache to a new runner); then the runners sleep for 10 minutes and when they woke up, they hit the limit again and go back to sleep.
Currently a runner does around ~60 req/min. With `1.5` as interval, the would decrease to ~34 req/min.

PR #161, opened 2018-02-27
[openclose_pr tool] Improving docs, logs and how rebase is done

felipevolpone:
Adding the docs of how to use openclose_pr.py script and how to deploy it. Also changing how the rebase is done before creating a new branch.


pvoborni:
Wouldn't it be better to split into 2 PRs? the two commits are not strictly related. Putting text to PR description would be also beneficial.

felipevolpone:
@pvoborni I think we are fine with just one PR, since we have two commits. I've updated the PR description.

PR #162, opened 2018-02-28
Do not use API call for downloading tasks file

slaykovsky:
Now downloading tasks file is using an GitHub API call
(GET /repos/:owner/:repo/pulls/:number/files). It can be useful
in some cases but not in ours. In our case the `.freeipa-pr-ci.yaml`
file exists either in the head or base branches. So we can retrieve
it just by constructing the URL. So then if the tasks file doesn't
exist in the head, we will just download it from the base.

### DIFF
--- a/github/prci_github/internals.py
+++ b/github/prci_github/internals.py
@@ -253,68 +253,37 @@ class Tasks(collections.Set, collections.Mapping):
         self.job_cls = job_cls
         self.tasks_conf = {}
 
-        for changed_file in pull.pull.files():
-            if changed_file.filename == tasks_config_path:
-                logger.debug('Tasks file was modified in PR %d. Using tasks '
-                             'file from PR.', pull.pull.number)
-                ref = pull.pull.head.sha
-                tasks_from_pr = True
-                break
-        else:
-            logger.debug('Task file was not modified in PR %d. Using tasks '
-                         'file from target branch', pull.pull.number)
-            ref = pull.pull.base.ref
-            tasks_from_pr = False
+        def log_retrieve(url):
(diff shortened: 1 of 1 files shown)


pvoborni:
@slaykovsky I recall that you didn't like the fact that a lot of previous commits in PR-CI doesn't have proper description or purpose. This PR lacks it as well.

stlaz:
@pvoborni Let me quote a part of our proposed Code of conduct:
‘’’
### Be careful in the words that you choose. 
We are a community of professionals, and we conduct ourselves professionally. Be
kind to others. Do not insult or put down other participants.
‘’’
Since you’re the manager of the FreeIPA project, I would expect you to follow it but I find you rubbing an assciate of yours’ words in his face rather offensive and unwelcoming.

felipevolpone:
Hi @slaykovsky, thank you for the patch :) 
Could you please add in the commit message the why the change is necessary and what it does? Now, it only says what it does not.

pvoborni:
I agree that I could write it in a better way. I'm sorry if it sounded harsh. My motivation is not to rub any word on anyone face. I'd like to increase the quality of our commit messages so that other developers who will work with the code base will have a way how to know why certain changes were done.  A lot of commit messages in this project doesn't have this explanation. If we want to change it then we should be better.

slaykovsky:
@felipevolpone Hi mate! I've updated the commit message, please take a look.

felipevolpone:
Hi @slaykovsky LGTM. I would like to propose one little change, could we have the `tasks_file_url` in the `logger.debug("Retrieving tasks file.")` call? This way it would improve debuggability :)

PR #164, opened 2018-03-12
Fix keyerror when trying to get tasks file

felipevolpone:
Previous code was raising `KeyError: 'repo'` on `tasks_file_url.format(url)`. We should only pass url as a parameter to session.get.


felipevolpone:
Merging it since it's a trivial fix.

PR #165, opened 2018-03-12
New Vagrant template: 1 master, 2 replicas

pvoborni:
The template was configured in a way that it can fit 3 IPA serves with
CA on runner with 8GB memory. Therefore replicas have bit less.

When thinking about it. I'm not sure if it would be better to set 2100MB for all 3 server - instead of, 2400, 2000 and 2000. Anyway testing with 2400 everywhere failed on third IPA installed as the runner got out of memory.

### DIFF
--- a/templates/vagrantfiles/Vagrantfile.master_2repl
+++ b/templates/vagrantfiles/Vagrantfile.master_2repl
@@ -1,0 +1,64 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+#
+
+Vagrant.configure(2) do |config|
+
+    config.ssh.username = "root"
+
+    config.vm.synced_folder "./", "/vagrant",
+        type: "nfs",
+        nfs_udp: false
+
+    config.vm.box = "{{ vagrant_template_name }}"
+    config.vm.box_version = "{{ vagrant_template_version }}"
+
+    config.vm.provider "libvirt" do |domain, override|
+        # Defaults for masters and replica
+        # WARNING: Do not overcommit CPUs, it causes issues during
+        # provisioning, when RPMs are installed
+        domain.cpus = 1
+        domain.memory = 2200
+
+        # Nested virtualization options
+        domain.nested = true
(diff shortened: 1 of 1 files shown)


pvoborni:
This is useful for: test_integration/test_topology.py::TestTopologyOptions as tried in https://github.com/pvoborni/freeipa/pull/8

slaykovsky:
Hi! I will put a link for a comment with similar thoughts as I have for this PR: https://github.com/freeipa/freeipa/pull/1506#issuecomment-372782250

akokshar:
Why do we need to have topology definition in advance? Wont it be better to provision a machine right when it is requested from the test?
Probably this was done like it is now because we had to know amount of necessary resources in advance in order to chose a proper runner - the one which is capable to host all VMs we need for a test. That is true, but why do we restricting ourselves to a single runner for a test? 
All IPA master/replicas/clients are independent. They only need a connectivity to talk to each other. 
We can create an overlay network, spanning all our runners. Then all VMs we provision would need to be attached to that virtual network. This way will need also to prevent different running tests from impacting each other but there can be plenty of ways to achieve this. 
My above proposal wont break an idea to have runners independent.

pvoborni:
@slaykovsky I agree, it could be generated. Having static topologies is a current limitation. On the other hand, we don't have that many topologies in our tests so it is not a big limitation and it is a matter of minutes to create a new one.

@akokshar IIRC the biggest motivation was to keep it as simple as possible - create something which works for our use cases. 

I don't think that this PR is the best place to discuss the proposals. You can create an RFE issue with the description, use case and value it should add.

felipevolpone:
Hi @pvoborni, do you have some runner running tests with this Vagrantfile? IIRC VMs with less memory (like using 2GB) could cause some failures.

pvoborni:
Following test run used the config: https://github.com/pvoborni/freeipa/pull/8 but it's true that it was tested only with a single test.

pvoborni:
This was an optimization. Might not be needed anymore thus closing.

PR #167, opened 2018-03-12
Export junit.xml in pytest to allow automatic processing

pvoborni:
We can easily produce also junit.xml so that system which can automatically
process it can do it from published artifacs. This might be one of the ways
how to automatically process nightly or other testing results.


slaykovsky:
LGTM

slaykovsky:
@felipevolpone pls see https://www.ibm.com/support/knowledgecenter/en/SSQ2R2_14.1.0/com.ibm.rsar.analysis.codereview.cobol.doc/topics/cac_useresults_junit.html

PR #168, opened 2018-03-12
Use GraphQL GitHub API alonside with REST

slaykovsky:
After we added more tests for PR to perform on CI, we're hitting
the GitHub REST API rate limit and overall performance of CI is
dramatically low.

Also, we have a problem that code which responsible for communication
with GitHub API is really hard to understand.

This patch is intended to fix these issues.

As GitHub also has GraphQL API endpoint which allows us to get all
the required information about every pull request in one API call
and its rate limit is different from REST API rate limit, this
patch is implementing the way of communicating with GitHub using
two GitHub API endpoints.

All the information gathering (think of it as GET requests) is
happening using GraphQL API. All the mutations of a state of pull
requests and commits statuses are happening through REST API, as
for now GraphQL API has really limited mutations support (think of it
as POST and PUT).

To achieve this way of APIs usage it was necessary to replace
one part of PR CI completely as it was intended to heavely use
github3py library in many places in the code.

Now there is World class which is responsible for communication
with outside world to change its state and a bunch of data classes
to map the data recieved from GitHub.

Also, I've added type hintings for a changed code so it will be
more easily for other contributors to understand this part of CI.

And in the end, I've wrote a bit of sanity unit tests for data
classes.

Auto update and auto restart of runners were intentionally removed
as for auto update it's bad in the first place to update infrastructure
without supervising and as for restart, nobody gave me the real reason
why it was implemented in the first place.


github/gql/h.py
```diff
@@ -0,0 +1,182 @@
...
+
+from requests import Session
+
+GITHUB_ENDPOINT = 'https://api.github.com/graphql'
+
+
+def make_headers(token):
+    """Make headers dict.
+
+    Args:
```
tomaskrizek: Documenting the functions' parameters is nice, but including the `Args` and `Returns` sections for each function seems a bit excessive. It's easy for these docstrings to get out of date and wrong documentation is usually worse than none.

I think [mypy(https://github.com/python/mypy)] is better suited for this purpose. Since PR CI is a standalone Python 3 project, it would be easy to use it. With mypy, you also get the benefits of static typing. The type checking could be automatically executed by Travis CI.

akokshar: @tomaskrizek, please feel free to submit a pull request. We will be happy to review it.

felipevolpone:
I haven't gone through the entire change/code, but I already see some needed changes:
* The PR lacks description
* The PR should be split into various commits, each of them should have a proper commit message. 
* As an example, I would create a separate commit to the addition of `firewalld` in `ansible/roles/runner/tasks/setup.yml` and explain in the commit message why it's necessary how it's related to the GraphQL change.
* `github/gql/h.py` is not a good name
* `github/gql/q.py` is not a good name
* I agree with @tomaskrizek about the `Args` and `Returns` sections, it's excessive and would easily be outaded.
* I see a lot of new code formatted as
```python
if (
    something here
) 
```

when 
```python
if (something here)
```
is how the rest of the code is formated.

* I would also like to see the this new change deployed in some machine and running some tests (in your own freeipa fork)

slaykovsky:
Hi!
1. It's not finished yet. So I will definetely add a description. 
2. Not for a graphql change, but very essential to have, because otherwise, it will not work on the ephemeral runner of some community guy with firewalld enabled. Will make another PR with this change.
3. Goto 2
4. The name is ok for me, will add a docstring there. If you think it's not a good one, please propose another one.
5. Goto 4
6. Didn't fully get your thought as `outaded` doesn't exist. For me, it does not matter whether to have this comments or crooked Python's types annotation.
7. Didn't get your point here.

Machines deployed, you can see my fork's PRs.

github/gql/h.py
pvoborni: IMO, the way how it is document is quite nice. Omitting the comments would be worse. And as you wrote Tomas, mypy is not currently used in the project. So it is a topic for other PR. There is no need to block this PR with mypy.

pvoborni:
I really like how entities and h.py and q.py are readable. Overall it's a good improvement from the previous state. In this comment I won't discuss details as I did not read them carefully. I also did not check if it remains the same feature wise.

That said I still think that all classes should have a doc string describing what they model and what is the purpose behind it, what they touch or what they should not touch.

The biggest issue in the code I see is modeling World as sort of singleton with ton of static methods.  Would rename it e.g. to GitHubRepo. This way it limits the implementation to be configured always against one repo and prevent the possibility to configure it e.g. against two. Use cases for 2 or more repos might be: 
* let's have one repo for nightly runs, second only for normal PRs so that we don't spam the development repo but can still use the same runners
* let's have a play-ground repo where we will send WIP patches without polluting the original repo while sharing infra
* let's configure the same runners against multiple projects to share lab resources E.g. SSSD team is considering using PR-CI. 

Usage of static methods makes it also harder to debug, to know what is set when. 

The second big problem I see is the main method. It is very long and contains 9 levels of indentation. This is hard read and orientate in. Separate substeps of it should be moved to function with only one purpose and no side effect. 

Let me paste here a concept (sort of python pseudo code) which illustrates the multiple repos idea in a way how PR-CI repo could be structured to not limit itself only to GitHUb - it also illustrates bots ideas - so take it as inspiration, not something that should be done by this pr as it is not finished. E.g. prioritzation of queues could be done differently.

```python
## PRCI-repo

'''
Contains functionality which is responsible for:
- getting new task
   - generating tasks (for some solutions)
- executing the tasks
- upload result/artifacts
- update itself

Is not responsible for launching vms and having implementation for
various tasks. That should be part of related project repo.

'''

class TaskDefinition:
    """
    Interface
    """

class TaskResult
    """
    Interface
    """


class JobsQueue:
    """
    Interface. Every source of jobs should implement this interface.
    """

    def get_config():
        """
        Loads config from runner config, section related to specific Queue
        """
        return dict()

    def next():
        """
        Get next task to run or None.
        """
        return TaskDefinition()

    def generate_tasks():
        """
        Generate new tasks

        Will be calls periodically. If wait is required then it should be
        controller but not waited in the Queue implementation.
        """
        return DateTime() # next event of generation

    def upload_result(task_result):
        # a way how to upload results of a task

    def upload_artifacts():
        # a way how to upload artifacts - this might or might not be
        # implemented. A task itself can do it.

    def download_artifacts(task):
        # the actual upload/download could be through configurable
        # provider, but at the beginning, simple rsync is OK.



class GitHubQueue(JobsQueue):
    """
    Getting tasks from new Git Hub Pull requests and control based on
    Git Hub status.
    """
    # override get_next(), generate_tasks()

class RabbitMQQueue(JobsQueue):
    """
    Getting new tasks from RabbitMQ
    """

class PRCI:

    queues: dict()
    config: dict()

    def init(self):
        self.config = load_config()
        # init logging
        self.initialize_queues()

    def initialize_queues(self):
        for queue in self.config.get('queues'):
            self.queues[name] = self.init_queue(queue)

    def run_task(self, task, queue):
        # prepare repo
        # create temp file with job definition
        # exec(repo/run_job --job $tempfile)
        # parse results
        queue.upload_result(results)
        queue.upload_artifacts()

    def run(self):
        while True:
            for queue in self.queues:
                task = queue.next()
                if task:
                    self.run_task(task, queue)
                queue.generate_tasks()
            wait(1)


## Project - repo

"""
Will contain
  * tasks (e.g. RunPytest, RunWebuiTests, Builder)
  * Ansible roles like: builder, machine
  * so basically also the Vagrant part

It will allow to run the same tests with the same method via other
executioners. E.g. as a part of Fedora or CentOS CI, basically every CI which
implements the standard test interface.
"""
```

felipevolpone:
@slaykovsky Hi mate, please, let me know when this PR is ready for review. I put the WIP label because you said this is still a WIP.

slaykovsky:
Hi @felipevolpone 

Sure thing!

pvoborni:
@slaykovsky  could you respond to the review comments here. It is not clear what is happening. You presented a Jenkins for a similar usage.  Does it mean that you want to abandon this PR? What is the goal of Jenkins work you are doing. E.g. could you the describe what problem are you solving or what approach you are investigating, how you want to proceed with this PR?

slaykovsky:
@pvoborni hi,

I will not abandon this PR, I will update it tomorrow.

> The second big problem I see is the main method. It is very long and contains 9 levels of indentation. This is hard read and orientate in. Separate substeps of it should be moved to function with only one purpose and no side effect.
> That said I still think that all classes should have a doc string describing what they model and what is the purpose behind it, what they touch or what they should not touch.

I completely agree with you.

> The biggest issue in the code I see is modeling World as sort of singleton with ton of static methods.

It has only 3 static methods and one of those is intended to be "private". :)

> it limits the implementation to be configured always against one repo

IIUC, now we need to have more runners to run tests with only GitHub usage. The other parts of your comment could be implemented in other patches.

It's too soon to talk about Jenkins stuff (for now I've implemented **only** TravisCI part), but today we've discovered a bug in 389-ds-base 1.4.0.6 on F28 while running xmlrpc tests (https://pagure.io/freeipa/issue/7456 https://github.com/freeipa/freeipa/pull/1714), so it's useful even in the state it now exists. :) The good part of Jenkins is I don't need to worry about infrastructure stuff much as it runs VM on demand on OpenStack. Then, the whole suit can be _theoretically_ defined only in one groovy script which will live in the freeipa repository. And finally, I don't think it's a good place to talk about Jenkins, especially right now.

Thanks.

ansible/roles/runner/tasks/deploy_pr_ci.yml
```diff
@@ -6,16 +6,6 @@
     version: "{{ pr_ci_repo_branch }}"
     force: true
 
-- name: install python dependencies with pip
```
felipevolpone: As was said in another PR we are not ready to drop installations by pip

ansible/roles/runner/tasks/setup.yml
```diff
@@ -26,11 +31,12 @@
     - createrepo
     - gzip
     - gcc
+    - firewalld
     - python3-devel
     - python3-pip
     - redhat-rpm-config
-    - redis
-    - python3-redis
+    - python3-github3py
```
felipevolpone: Let's stay with pip

ansible/roles/runner/tasks/setup.yml
```diff
@@ -39,9 +45,29 @@
...
+  firewalld:
+    service : "{{ item }}"
+    permanent: true
+    state: enabled
+  with_items:
+    - nfs
+    - ipp
+    - rpc-bind
+    - mountd
+  notify: firewalld complete reload
```
felipevolpone: This was addressed by another PR right? so it should not be on this on I guess

slaykovsky:
@felipevolpone 
one https://github.com/freeipa/freeipa-pr-ci/pull/168/files#diff-fd06f0e528b175b32e8b53cdf77591a8R1
two https://github.com/freeipa/freeipa-pr-ci/pull/168/files#diff-420b6c870dff05aff56160dcd9599d44R1

felipevolpone:
It's really cool that we now have real tests, but I'm not sure how to run them. Could you please update developer's doc to cover this part?

slaykovsky:
@felipevolpone sure

ansible/roles/runner/tasks/setup.yml
slaykovsky: Okay

ansible/roles/runner/tasks/setup.yml
slaykovsky: While another one is not merged I need to have it somewhere.

Will remove this part later.

pvoborni:
>> The biggest issue in the code I see is modeling World as sort of singleton with ton of static methods.

> It has only 3 static methods and one of those is intended to be "private". :)

It is more than three. Other classes have also static methods which depend on worlds behavior. It behaves as one organism - a change in one will force a change in other.

Do you agree with a proposal to change it from static world to a set of a class which can have proper instances?  

>> it limits the implementation to be configured always against one repo

>  IIUC, now we need to have more runners to run tests with only GitHub usage. The other parts of your comment could be implemented in other patches.

I agree, not a topic for this PR, as I wrote in `it also illustrates bots ideas - so take it as inspiration, not something that should be done by this pr `. But it is important to mention in order to preven design decisitons which would make future enhancements more difficult.

> It's too soon to talk about Jenkins stuff (for now I've implemented only TravisCI part).

The usual, preffered way in any bigger work is to first propose the idea. Explain what are the benefits and added value in comparison to an existing solution. Then ask for feedback. If possible, best to do it publicly, in our case freeipa-devel list.  If feedback is positive then a Proof of Concept can be scheduled in following sprint. This  is important to check to know if any actual work should even start.  So no, it is not too soon. Actually it is too late now.

You jumped right into implementation with undefined scope and purpuse. To my understanding, rest of the team, maybe except people who sits next to you, have no idea what you are doing.  You risk that the effort you put into it will be simply discarded.

> And finally, I don't think it's a good place to talk about Jenkins, especially right now.

It is not a good place to discuss the solution here. But the work is related to this PR as it consumes your time and thus slows completion of this PR. Doing parallel work also leads to lower efficiency. In fact, IMO, more people should work on this PR too speed it up, have quicker feedback loop.  Is the Jenkins work more important than this PR? If so then explain why. If not then let's focus on this one.

Our PR testing is drastically slowed now. It is important to unblock it so that entire team doesn't suffer from the slowness.

slaykovsky:
BTW, I forgot about whitelist here.

stlaz:
@pvoborni please stop sidetracking this PR discussion and keep it relevant to these code changes.

pvoborni:
We had an offline discussion. I won't discuss the Jenkins part here any longer.

On the singleton side. I agree that we don't need to change the PR to support more instances in this PR and focus on unblocking upstream testing.  But keep possible enhancements in mind when creating new internals of the project. 

This PR suffers from not getting good enough review. Due to other duties, I can only do high-level review. @felipevolpone  could do the whole, but this and previous sprint he has other focus which takes a majority of his time. @tiran, could you work with @slaykovsky on this till the push? I know that your understanding of PR-CI code is limited, but that is not that different from other team members. This change is a complete rewrite of a part of the codebase so previous inexperience should not limit you much. I think that your input will be valuable here. I'm sure that @felipevolpone or @slaykovsky will answer any question you will have.

slaykovsky:
Bug: it's not catching github3.exceptions.NotFoundError when removes a label which doesn't exist.

github/prci.py
```diff
@@ -4,140 +4,37 @@
...
-
-        self.done = True
-        self.aborted = True
+import yaml
 
-    def quit(self):
-        logger.info("Quiting just now! No results will be reported.")
-        sys.exit()
+import internals.entities as entities
+from internals.gql import h, q
```
felipevolpone: As I said about the file names, I would prefer to have variables with a real name instead of just a char.

github/internals/entities.py
```diff
@@ -0,0 +1,799 @@
...
+        self.state.session = session
+        self.state.repo_owner = repo_owner
+        self.state.repo_name = repo_name
+        self.state.runner_id = runner_id
+        self.state.tasks_path = tasks_path
+        self.state.whitelist = whitelist
+        self.state.instance = self
+
+    @staticmethod
+    def __check_limit(which=None):
```
felipevolpone: Maybe we could change `which` for `protocol`

github/prci.py
```diff
@@ -247,49 +210,82 @@ def main():
...
+        if not isinstance(response, dict):
+            logger.error(response)
+            sys.exit(1)
+
+        data = h.get_data(response)
+        repository = h.get_repository(data)
+        repository_url = h.get_repository_url(repository)
+        pull_requests_data = h.get_pull_requests(repository)
+
+        pull_requests = sorted(
```
felipevolpone: This indentation looks weird to me. It doesn't seem to be as it is in other parts of the project.

Maybe this looks better, not sure
```python
pull_requests = sorted((entities.PullRequest.from_dict(pr_data)
                        for pr_data in pull_requests_data),
                        key=lambda pr: not pr.prioritized)
````

File: github/prci.py [hunk 8]
```diff
@@ -187,57 +84,123 @@ def load_yaml(yml_path):
...
-            f.write(str(next_reboot))
+def process_pull_request(pull_request, repository_url):
+
+    if pull_request.postponed:
+        skipping_pr("postponed.", pull_request.number)
+        return None
+
+    if not pull_request.mergeable:
+        skipping_pr("can't be merged.", pull_request.number)
+        if not pull_request.needs_rebase:
```
felipevolpone: Actually, is without the `not`, isn't it? If the PR cannot be merged and needs rebase, it adds the label.

slaykovsky: No, it's actually not https://github.com/freeipa/freeipa-pr-ci/pull/168/files#diff-d4101ce55c387d10a96f9a7088cbf8edR417

doc/README.md
```diff
@@ -404,15 +389,15 @@ available, this limit could be hit.
 
 #### Abuse limit
 
-Even the cached requests (wich return 304) are limited by the GitHub API.
+Even the cached requests (which return 304) are limited by the GitHub API.
```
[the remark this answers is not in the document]
slaykovsky: But still, CacheControl is being used here https://github.com/freeipa/freeipa-pr-ci/blob/master/github/open_close_pr.py

github/internals/entities.py
```diff
@@ -0,0 +1,799 @@
...
+
+    def register_task(self, task):
+        self.task = task
+
+    def unregister_task(self):
+        self.task = None
+
+
+class AbstractJob(Callable):
+    __metaclass__ = ABCMeta
```
stlaz: `python3` disregards `__metaclass__` and still considers this class to be concrete unless it inherits from `ABC`:
`python3`:
```python
>>> class AbstractJob():
...        __metaclass__ = ABCMeta
...        def __init__(self, a, b):
...                self.a = a
...                self.b = b
...        @abstractmethod
...        def x(self):
...            return self.a, self.b
...
>>>
>>> x = AbstractJob()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: __init__() missing 2 required positional arguments: 'a' and 'b'
>>> x = AbstractJob(1,2)
>>> x
<__main__.AbstractJob object at 0x7f12ad2e15c0>
```
whereas `python2`:
```python
>>> class AbstractJob():
...        __metaclass__ = ABCMeta
...        def __init__(self, a, b):
...                self.a = a
...                self.b = b
...        @abstractmethod
...        def x(self):
...            return self.a, self.b
...
>>> x = AbstractJob()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: Can't instantiate abstract class AbstractJob with abstract methods x

```

github/internals/entities.py
```diff
@@ -0,0 +1,799 @@
...
+            else:
+                raise
+        else:
+            state = State.SUCCESS
+            description = ''
+
+        return JobResult(state, description, url)
+
+
+class JobDispatcher(AbstractJob):
```
stlaz: Is this still an Abstract Base Class for something or is it concrete?

github/internals/entities.py
```diff
@@ -0,0 +1,799 @@
...
+    def __init__(self, state, description="", url=""):
+        if state not in self.valid_states:
+            raise ValueError('invalid state: {}'.format(state))
+
+        self.state = state
+        self.description = description[:GITHUB_DESCRIPTION_LIMIT]
+        self.url = url
+
+
+class Job(AbstractJob):
```
stlaz: Is this still an abstract class?

doc/README.md
```diff
@@ -4,7 +4,7 @@
 
 ![overview](images/overview.svg)
 
-A swarm of independent machines called *runners* is monitoring Pull Requests
+A swarm of independent machines called *runners* are monitoring Pull Requests
```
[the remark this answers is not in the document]
slaykovsky: Yup, sorry :)

ansible/roles/runner/tasks/setup.yml
```diff
@@ -26,22 +31,33 @@
...
-    - python3-parse
-    - python3-dateutil
-    - python3-psutil
-    - python3-PyYAML
-    - python3-raven
-    - python3-tqdm
-    - python3-pytz
-    - python3-jinja2
+    - python-firewall
+    - python3-firewall
```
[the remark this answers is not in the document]
slaykovsky: I've added in in `requirements.txt` as we are intended to use pip to install all the dependencies.

github/prci.py
stlaz: `util` and `queries` maybe?

doc/README.md
```diff
@@ -473,3 +450,11 @@ Then run:
...
 ansible-playbook -i ansible/hosts/runners ansible/prepare_openclose_pr_tool.yml
 ```
+
+## How to run available unit tests
+Simple in a project root run
+```
+pytest-3.6
+```
+
+Note that you need to have all required dependencies installed.
```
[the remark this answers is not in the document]
slaykovsky: Added `requirements.txt`

github/internals/entities.py
```diff
@@ -0,0 +1,799 @@
...
+@unique
+class State(CIEnum):
+    """Enum which represents the status of a commit"""
+    PENDING = "PENDING"
+    FAILURE = "FAILURE"
+    SUCCESS = "SUCCESS"
+    ERROR = "ERROR"
+
+
+class Map(dict):
```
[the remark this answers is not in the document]
slaykovsky: Deleted

github/prci.py
slaykovsky: `util` and `queries` it is

github/internals/entities.py
slaykovsky: Changed to `resource`

ansible/roles/runner/tasks/setup.yml
```diff
@@ -39,9 +43,29 @@
     - python3-tqdm
     - python3-pytz
     - python3-jinja2
+    - python-firewall
+    - python3-firewall
```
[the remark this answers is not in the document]
slaykovsky: I've added `requirements.txt`

github/internals/entities.py
```diff
@@ -0,0 +1,799 @@
...
+EPHEMERAL_LIMIT = 60
+STALE_TASK_EXTRA_TIME = 60
+
+
+def sentry_report_exception(context=None):
+    sentry = raven.Client(SENTRY_URL)
+
+    if context is not None:
+        with sentry.context:
+            sentry.context.merge(context)
```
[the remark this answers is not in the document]
slaykovsky: Done :)

github/internals/entities.py
```diff
@@ -0,0 +1,799 @@
...
+
+GITHUB_DESCRIPTION_LIMIT = 139
+RACE_TIMEOUT = 17
+RERUN_PENDING = "pending for rerun"
+TASK_TAKEN_FMT = "Taken by {runner_id} on {date}"
+SENTRY_URL = (
+    "https://d24d8d622cbb4e2ea447c9a64f19b81a:"
+    "4db0ce47706f435bb3f8a02a0a1f2e22@sentry.io/193222"
+)
+EPHEMERAL_LIMIT = 60
```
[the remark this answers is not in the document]
slaykovsky: Added a comment.

github/internals/entities.py
```diff
@@ -0,0 +1,799 @@
...
+    """Enum which represents the status of a commit"""
+    PENDING = "PENDING"
+    FAILURE = "FAILURE"
+    SUCCESS = "SUCCESS"
+    ERROR = "ERROR"
+
+
+class Map(dict):
+    """dot notation access to dictionary attributes"""
+    __getattr__ = dict.get
```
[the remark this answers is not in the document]
slaykovsky: I've deleted `Map`

github/internals/entities.py
```diff
@@ -0,0 +1,799 @@
...
+            limit=data_dict["limit"],
+            remaining=data_dict["remaining"],
+            reset_at=data_dict["reset"]
+        )
+
+    @staticmethod
+    def __get_limit(which=None):
+        """Calls GitHub API and returns RateLimit instance"""
+        if which not in ["graphql", "core"]:
+            return None
```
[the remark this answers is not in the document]
slaykovsky: I've updated the patch, still not sure if it's good enough :)
```python
    def get_rate_limit(self, resource: Text=None) -> RateLimit:
        """Calls GitHub API and returns RateLimit instance"""
        if resource not in RateLimit.valid_resources:
            ValueError("Supported resources are graphql and core")

        return RateLimit.from_dict(
            self.github_api.rate_limit()["resources"][resource]
        )
```

github/internals/entities.py
```diff
@@ -0,0 +1,799 @@
...
+        """Calls GitHub API and returns RateLimit instance for GraphQL"""
+        return RateLimit.__get_limit("graphql")
+
+
+class World(object):
+    """Represents the outside world state"""
+    state = Map()
+
+    def __init__(
+        self, graphql_request, github_api, session, repo_owner,
```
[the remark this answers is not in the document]
slaykovsky: Done

github/internals/entities.py
```diff
@@ -0,0 +1,799 @@
...
+        now = datetime.now(pytz.UTC)
+        timeout = timedelta(seconds=task.timeout)
+        if not timeout:
+            return False
+        parsed = parse.parse(TASK_TAKEN_FMT, self.description)
+        if not parsed:
+            return False
+
+        taken_on = parser.parse(parsed["date"])
+        # TODO: Extra time???
```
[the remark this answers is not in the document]
slaykovsky: Nope :)

slaykovsky:
Hi @stlaz  and @felipevolpone 

I've updated the patch. Pls, do the review once more :)

P.S. Many kudos to @stlaz

github/internals/entities.py
```diff
@@ -0,0 +1,768 @@
...
+            state = State.ERROR
+            sentry_report_exception({"module": "tasks"})
+        else:
+            description = job.description
+            if job.returncode == 0:
+                state = State.SUCCESS
+            else:
+                state = State.FAILURE
+        finally:
+            return JobResult(state, description, job.remote_url)
```
stlaz: I don't think this particular return should be in `finally` block as it would eat any exception raised in `except`/`else` blocks (although there aren't many points where that could happen right now). This was quite interesting for me to learn, actually.
Try running the following snippet in the python interpreter if you want:
```python
import sys

def x():
        try:
                raise RuntimeError('something')
        except Exception as e:
                raise RuntimeError('something2')
        finally:
                return 5

print(x())
print(sys.exc_info())
```
The effort to really run the `return` in the `finally` block is so high that the raised exception is eaten.

github/internals/gql/queries.py
```diff
@@ -0,0 +1,79 @@
...
+      }
+    }
+  }
+  rateLimit {
+    limit
+    cost
+    remaining
+    resetAt
+  }
+}""" % (owner, repo)}
```
stlaz: Please use `str.format()` instead.

github/internals/gql/queries.py
```diff
@@ -0,0 +1,79 @@
...
+      }
+    }
+  }
+  rateLimit {
+    limit
+    cost
+    remaining
+    resetAt
+  }
+}""" % (owner, repo, pr_number)}
```
stlaz: Please use `str.format()` instead.

github/internals/entities.py
```diff
@@ -0,0 +1,768 @@
...
+            test_suite: path/to/test.py
+            template: *ci - master - f27
+            timeout: 3600
+            topology:
+                name: topology_name
+                cpu: 4
+                memory: 5750
+        """
+        self.task_class = getattr(
+            __import__('tasks', fromlist=[job_data['class']]),
```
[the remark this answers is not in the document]
slaykovsky: Hi! Actually, we don't need to have `import_module` here as we can import `tasks` module from `tasks` package and just call a `getattr` function :)

```python
...
from tasks import tasks
...
self.task_class = getattr(tasks, job_data["class"])
```

github/internals/gql/queries.py
slaykovsky: But it would not work in this case...
```python
>>> "{ bla { blabla { {bar} } } }".format(bar="bar")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: unexpected '{' in field name
```
And to make it work any curly braces which are not the part of formatting placeholder:
```python
>>> "{{ bla {{ blabla {{ {bar} }} }} }}".format(bar="bar")
'{ bla { blabla { bar } } }'
```

I really don't mind to change it if you're ok with doubling :)

github/internals/entities.py
```diff
@@ -0,0 +1,768 @@
...
+        """Constructs the instance of a Task to be processed by the handler"""
+        self.name = name
+        self.pr_number = pr_number
+        self.commit_sha = commit_sha
+        self.repo_url = repo_url
+        self.ref_spec = "pull/{}/head".format(pr_number)
+
+        # The data should be a dictionary with the following structure:
+
+        # fedora - 27 / simple_replication:
```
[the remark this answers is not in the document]
slaykovsky: Nope

github/internals/entities.py
```diff
@@ -0,0 +1,768 @@
...
+
+    def check_graphql_limit(self) -> None:
+        return self.__check_limit("graphql")
+
+
+class Topology(object):
+    def __init__(
+        self, name: Text=None, memory: SupportsFloat=None, cpu: int=None
+    ) -> None:
+        if name is None:
```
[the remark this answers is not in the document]
slaykovsky: Thanks, done

github/internals/entities.py
```diff
@@ -0,0 +1,768 @@
...
+                owner=world.repo_owner,
+                repo=world.repo_name,
+                path=world.tasks_path,
+                sha=self.commit.sha
+            )
+        )
+        if res.status_code != 200:
+            # If the file doesn't exist in the commit, we'll get in from the
+            # base branch
+            # Actually, this branch should never be executed...
```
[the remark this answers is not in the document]
slaykovsky: No, as the contributor may have an old code base on which they did changes so the tasks file could be not in this branch and in this case we will get the tasks file from a base branch.

github/internals/entities.py
```diff
@@ -0,0 +1,768 @@
...
+
+class Topology(object):
+    def __init__(
+        self, name: Text=None, memory: SupportsFloat=None, cpu: int=None
+    ) -> None:
+        if name is None:
+            name = "undefined"
+        if memory is None:
+            memory = AvailableResources.initial_memory
+        elif not isinstance(memory, float):
```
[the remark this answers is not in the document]
slaykovsky: Okay

github/internals/entities.py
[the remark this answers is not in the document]
stlaz: Did not think of that, good thinking.

github/internals/entities.py
```diff
@@ -0,0 +1,758 @@
...
+        #   priority: 50
+        #   job:
+        #       class: RunPytest
+        #       args:
+        #           build_url: '{fedora-27/build_url}'
+        #           test_suite: test_integration / test_simple_replication.py
+        #           template: *ci-master-f27
+        #           timeout: 3600
+        #           topology: *master_1repl
+        data = defined_tasks[name]
```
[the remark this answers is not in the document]
slaykovsky: Thanks!

github/prci.py
```diff
@@ -187,109 +83,219 @@ def load_yaml(yml_path):
...
-        sentry.captureException()
-    finally:
-        sentry.context.clear()
+        tasks_data = pull_request.get_tasks_data(world)
+    except (yaml.error.YAMLError, TypeError, KeyError) as e:
+        logger.error(e)
+        return None
+
+    if pull_request.needs_rerun:
+        if all(map(
```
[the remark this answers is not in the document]
slaykovsky: It's more like `if all statuses are not failed and re-run label was set previously, remove the re-run label` :)

github/prci.py
```diff
@@ -247,49 +217,80 @@ def main():
...
+        runner_id=runner_id,
+        tasks_path=tasks_path,
+        whitelist=whitelist
+    )
 
-if __name__ == '__main__':
+    while not exit_handler.done:
+        world.check_graphql_limit()
+
+        response = do_request(
```
stlaz: `EnvironmentError` may be raised. Should it be treated?

github/prci.py
```diff
@@ -247,49 +217,80 @@ def main():
...
+            query=queries.make_pull_requests_query(
+                world.repo_owner, world.repo_name
+            )
+        )
+
+        if not isinstance(response, dict):
+            logger.error(response)
```
stlaz: `logger.error("Expected github response to be an instance of type 'dict', but it is '%s'. The response: %r", type(response), response))` or split that into two `logger.error()` calls.

github/prci.py
slaykovsky: Addressed :)

github/prci.py
slaykovsky: Removed

github/prci.py
```diff
@@ -247,49 +217,80 @@ def main():
...
+                world.available_resources.take(task)
+                logger.info(
+                    "Available resources: %s", world.available_resources
+                )
+                try:
+                    task.execute(world, pull_request.commit.statuses)
+                except (EnvironmentError, RuntimeError) as e:
+                    logger.error(e)
+                    sentry_report_exception({"module": "github"})
+                    sleep(ERROR_BACKOFF_TIME)
```
felipevolpone: I think we can get rid of this. IIUC the purpose of this in the first place was to avoid hitting the API rate limit.

github/internals/entities.py
```diff
@@ -0,0 +1,757 @@
...
+        if resource not in RateLimit.valid_resources:
+            ValueError("Supported resources are graphql and core")
+
+        return RateLimit.from_dict(
+            self.github_api.rate_limit()["resources"][resource]
+        )
+
+    def poll_status(
+        self, pr_number: int, task_name: Text
+    ) -> "Status":
```
felipevolpone: Shouldn't we use postponed annotation evaluation instead of strings?
https://www.python.org/dev/peps/pep-0563/

felipevolpone:
I have juts two comment, the rest LGTM.
- The commit message has: `limited mutations support (this of it as POST and PUT)`. I guess that is `think` instead of `this`.
- The documentation on `doc/README.md` about how to run the tests is outdated. It should be updated to reflect the `./run-pytest`. Once I see it updated, I'll run the tests and if they are green, I will merge this PR.

@slaykovsky What's your plan to deploy the runners? I think we could deploy two or three runners with this new code, monitor them for a week or two and them change the rest. What do you think?

github/prci.py
slaykovsky: But we will reach the rate limit very quickly if we will not wait for some time anyway.

slaykovsky:
Hi @felipevolpone 

> The commit message has: limited mutations support (this of it as POST and PUT). I guess that is think instead of this.

Thanks!

> The documentation on doc/README.md about how to run the tests is outdated. It should be updated to reflect the ./run-pytest. Once I see it updated, I'll run the tests and if they are green, I will merge this PR.

But I didn't change `run-pytest` and this change doesn't affect it's code in any way, so I just don't know what you want to do :)

> What's your plan to deploy the runners?

I don't have any plan but it could be deployed right away. If you want to give it a try first, I have no problem with it.

github/internals/entities.py
slaykovsky: If you can point what should I change it will be great because in this PEP I see it's about Python 4.0 and Python 3.7

felipevolpone: ah pff, forget what I said.

pvoborni:
Commit message says:

> Auto update and auto restart of runners were intentionally removed
as for auto update it's bad in the first place to update infrastructure
without supervising

What is the plan for updating the runners over time - for additional PR-CI changes, after this PR is merged? Right now it is done automatically which doesn't require manual work. After merging this it will need to be done by somebody. How will be the "update even" chosen? 

> and as for restart, nobody gave me the real reason
why it was implemented in the first place.

```
git log --oneline -200 | grep reboot
git show 0836fde
```
Are the conditions explained in the commit message of 0836fde2b73bac4c48731de7c4d0a5b91bbe2e17 gone?

slaykovsky:
Hi @pvoborni 

> What is the plan for updating the runners over time - for additional PR-CI changes, after this PR is merged? Right now it is done automatically which doesn't require manual work. 

As this CI is a blocking point for the whole team it's really strange to make it update itself. There's a lot of things that can go wrong. I'll give you several of examples:

* repository can't be reached for some reason
* someone pushes not working properly code by mistake
* someone intentionally pushes code that brakes everything

It will lead to the error state when the whole system doesn't work.

> After merging this it will need to be done by somebody.

I believe that everyone in the team has skills to run ansible playbook against the given inventory.

>  How will be the "update even" chosen?

The update should be executed when the system in the almost standby state. We can choose the day of the week.

> Are the conditions explained in the commit message of 0836fde gone?

As I wrote in the commit message:
> nobody gave me the real reason

I don't see any cause in the commit message that will give me any useful information. Reboot is not the solution. I didn't see any uncontrollable performance degradation on my machines except for cases when I've killed prci python process and there were several VMs which were not killed and because of it swap was full and perf showed me that swapper is eating all the time. And this is also a problem that should be fixed properly.

pvoborni:
The auto update is a feature which can be used by other people. E.g. one can have a branch which is considered stable and runner is configured against it. People might rely on it.

I'm not against not having the functionality. But it would be good to communicate the change at least on devel list to prepere for the change and not be surprised that the runner is not updating itself.

Wrt reboot - if you - @felipevolpone  and @slaykovsky , agree that it is no longer an issue - PR-CI works long term fine and the workaround is not needed, then OK.

tomaskrizek:
If you want to remove the periodic reboot, just make sure your hosts are running the most recent Fedora with updated packages. The underlying issue might've been fixed in the mean time. If not, you'll definitely be able to tell, because tasks will start to time out after a few hours of runner uptime.

stlaz:
@felipevolpone I have no idea what you mean by "team values" but in the argument "what's better - unicorns or lollipops", I will be rooting for lollipops simply for the fact that they are real and can be really tasty.

But really - there's over 20 people who can push to this repository at their will and thus more than 20 subjects vulnerable to social engineering/being hacked. It does not even need to be intentional by the actual person in our team.

As for the crashes - I believe no such scenarios have been observed lately with this PR's code and #180 should greatly decrease possible resources leak.

stlaz:
We agreed on IRC that this could be pushed.

PR #169, opened 2018-03-14
Ensure firewalld is installed and configured

slaykovsky:
Now runner preparation does not configure firewalld service and
open ports for NFS related services. As the result Vagrant can't
mount NFS directory so no task will be run.

This patch eliminates this kind of thing.


felipevolpone:
I never faced such error or issues. Could you explain it to me (or give some example)?

felipevolpone:
Ah now I got it. Thanks @slaykovsky ! The patch LGTM

PR #170, opened 2018-03-14
Install Python packages from Fedora repo

slaykovsky:
Now almost every Python package that are dependencies for runner's
code getting installed from Fedora repository, except of github3 and
CacheControl.

This patch removes pip usage and makes ansible to install required
packages from the Fedora repo.

### DIFF
--- a/ansible/roles/runner/tasks/deploy_pr_ci.yml
+++ b/ansible/roles/runner/tasks/deploy_pr_ci.yml
@@ -6,16 +6,6 @@
     version: "{{ pr_ci_repo_branch }}"
     force: true
 
-- name: install python dependencies with pip
-  pip:
-    executable: pip3
-    state: latest
-    name: "{{ item.name }}"
-  with_items:
-    - name: "setuptools"
-    - name: "github3.py>=1.0.0a4"
-    - name: "CacheControl>=0.12.3"
-
 - name: create config directory if it dosn't exist
   file:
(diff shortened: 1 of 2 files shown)


ansible/roles/runner/tasks/deploy_pr_ci.yml
```diff
@@ -6,16 +6,6 @@
...
 
-- name: install python dependencies with pip
-  pip:
-    executable: pip3
-    state: latest
-    name: "{{ item.name }}"
-  with_items:
-    - name: "setuptools"
-    - name: "github3.py>=1.0.0a4"
-    - name: "CacheControl>=0.12.3"
```
tomaskrizek: There is a good reason the minimal version of CacheControl is `>=0.12.3`. I don't remember the details, but there were some intricate caching issues.

Until Fedora 28, you have two options: get it from pip, or use my copr repo [tkrizek/freeipa-pr-ci](https://copr.fedorainfracloud.org/coprs/tkrizek/freeipa-pr-ci/) (this is how it used to be).

Tracked bug for rebase: https://bugzilla.redhat.com/show_bug.cgi?id=1484313

felipevolpone:
Yeah, TBH I also don't remember why we need > 0.12.3. So it's just safer stay with pip (since fedora doesn't provide it).

slaykovsky:
@tomaskrizek @felipevolpone thank you for pointing to this one. Closing.

PR #171, opened 2018-03-14
Upgrade runner's packages during its setup

slaykovsky:
Now the runner's packages are not getting upgraded and this leads
to many problems, such as security issues or not working executables.

This patch adds dnf upgrade call during runner's setup.


felipevolpone:
LGTM.

PR #172, opened 2018-03-14
Make logger format fancy

slaykovsky:
Now logger format doesn't contain spaces after its level name,
module name and the message.

This patch adds spaces so log becomes a little easier to read.


felipevolpone:
LGTM

PR #173, opened 2018-03-19
test_webui: add host and service csr

Rezney:
Add host and service csr in order to test the currently skipped
certificate related tests.


felipevolpone:
Hi @Rezney thank you for your PR! I'll review as soon as I can.

pvoborni:
Look good to me but I have not tested it so have no idea if it works.

Rezney:
@felipevolpone Thanks for running this PR. I realized that we are generating host cert for different host (obviously:). So I fixed the issue we got in the test run. Apart from that I also added user crt.

Please re-run the test on your side when you have some time.

Cheers!

felipevolpone:
Tested and works perfectly! Thank you very much for this PR @Rezney :100: :confetti_ball:

PR #176, opened 2018-03-23
Add ppicka to whitelist

pavelpicka:
Signed-off-by: Pavel Picka <ppicka@redhat.com>

### DIFF
--- a/whitelist.yml
+++ b/whitelist.yml
@@ -25,6 +25,7 @@
 - mzidek-rh
 - ndehadrai
 - npmccallum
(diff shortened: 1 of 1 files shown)


whitelist.yml
```diff
@@ -37,3 +37,4 @@
 - tiran
 - tomaskrizek
 - tscherf
+- pavelpicka
```
tomaskrizek: `diff -u whitelist.yml <(sort whitelist.yml)` :cry:

PR #177, opened 2018-03-25
Artifacts: do not compress screenshots (png files)

pvoborni:
Screenshot is a png - already compressed. GZIP compression doesn't save much
additional space.

Compressing screenshots creates UX issue: Firefox is not able to directly
open png.gz file. So one must download it, uncompress and then open which
takes time.

By not compressing screenshots, one can easily view Web UI test failure
directly from uploaded artifacts.

### DIFF
--- a/tasks/remote_storage.py
+++ b/tasks/remote_storage.py
@@ -17,6 +17,7 @@ class GzipLogFiles(PopenTask):
             '-a ! -path "*/assets/*" '
             '-a ! -path "*/rpms/*" '
             '-a ! -name "*.gz" '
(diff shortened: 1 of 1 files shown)


pvoborni:
Example run at https://github.com/pvoborni/freeipa/pull/11 resp. artifacts: at https://fedorapeople.org/groups/freeipa/prci/jobs/810d8b2a-3024-11e8-be11-fa163e1a89da/

felipevolpone:
LGTM.
However, I wonder if we should "change" what PR 166 did, and create a folder to put the screenshot, instead of having them on the root dir.

PR #180, opened 2018-04-04
tasks/vagrant.py VagrantCleanup fix

slaykovsky:
Now the fallback after Vagrant destroy PopenException is completely
useless as it will not kill Vagrant and will not restart libvirtd
service.

This patch fixes it.

Also added pkill call to make sure there are no lost virtual machines.


felipevolpone:
LGTM

slaykovsky:
@felipevolpone @stlaz thanks, merging.

PR #183, opened 2018-04-05
Fix poll_status to raise EnvironmentError

slaykovsky:
Now poll_status method of a World class throws the RuntimeError
which is not being catched so the whole PR CI instance dies.

This patch fixes it.


felipevolpone:
Tested and it fixes the problem

PR #184, opened 2018-04-05
Task.lock: don't lock already processed task

slaykovsky:
Now the already processed (succeeded and failed) task will be
locked if it was gathered and yielded by prci before it was
taken by any runner and changed status from pending.

This patch adds the check for this kind of thing to the lock
method of Task class and fixes the behaviour.


felipevolpone:
LGTM

PR #187, opened 2018-04-06
Deployment: restart the runners after deploy

slaykovsky:
As we now have the packages upgrade process running during deployment,
and auto update is eliminated, we need to reboot the runner after
deployment complete to pick up new kernel, libraries, prci code changes.

This patch adds reboot to the post tasks of deployment.

### DIFF
--- a/ansible/prepare_test_runners.yml
+++ b/ansible/prepare_test_runners.yml
@@ -17,3 +17,17 @@
       deploy_ssh_key: true
       enable_nested_virt: true
       create_systemd_unit: true
+  post_tasks:
+    - name: reboot host
+      shell: sleep 2 && /sbin/shutdown -r now "Packages upgrade"
+      async: 1
+      poll: 0
+      ignore_errors: true
+    - name: wait for server to come back
+      local_action: wait_for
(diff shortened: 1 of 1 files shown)


felipevolpone:
Hi @slaykovsky, I don't think that this approach is doable. I tried to run only one time with just one runner and it already failed with a timeout error:
```
TASK [reboot host] ****************************************************************************************************************************************************************************
changed: [ipa-prci-01]

TASK [wait for server to come back] ***********************************************************************************************************************************************************
fatal: [ipa-prci-01 -> localhost]: FAILED! => {"changed": false, "elapsed": 300, "msg": "Timeout when waiting for ipa-prci-01:22"}

msg: Timeout when waiting for ipa-prci-01:22
        to retry, use: --limit @/home/fbarreto/projects/freeipa-pr-ci/ansible/prepare_test_runners.retry

PLAY RECAP ************************************************************************************************************************************************************************************
ipa-prci-01                : ok=30   changed=11   unreachable=0    failed=1

```

I think we could remove the task `wait for server to come back` and go only with the reboot.

PR #195, opened 2018-04-19
Cleanup after reprovision of VM fails

slaykovsky:
Now __setup_provision procedure reprovisions a box another time
after the first try failed. If provision will fail again it would
leave machines up and running, which will lead to many nasty things.
This happens because PopenException is not being catched in
with_vagrant wrapper.

This patch makes wrapper to catch PopenException to eliminate this
behaviour.

### DIFF
--- a/tasks/vagrant.py
+++ b/tasks/vagrant.py
@@ -9,7 +9,7 @@ def with_vagrant(func):
     def wrapper(self, *args, **kwargs):
         try:
             __setup_provision(self)
-        except TaskException as exc:
+        except (PopenException, TaskException) as exc:
             logging.critical('vagrant or provisioning failed')
             raise exc
         else:


tomaskrizek:
This used to solve a lot of false-negative provisioning issues. I remember there were various transient issues, from failed nfs provisioning, to various network issues, and this helped us to filter a lot of them.

pvoborni:
The issue is that there are leftovers after a second fail. So why not clean it up again after the failure as well?  The time the CI system spends on a retry seems better spent then human investigation of a flake in another case.

pvoborni:
The change looks good to me.

stlaz:
Yet `PopenException` is a subclass of `TaskException` so the proposed change should not really be changing anything.

slaykovsky:
Thanks, @stlaz !

The underlying issue seems to be somewhere else. I will close this PR and will continue to find the real issue.

Issue #194, opened 2018-04-18
Failure after PR is forced pushed when test was running


pvoborni:
When PR is forced pushed when test is running. The test will be finished and then PR-CI will error out on status update step with 
```
ERROR: __main__: Can't parse status data.
```
It is expected as the force push removed the status. But it causes upload of data to sentry.io. 

Given that it is a fairly common thing then this situation should be handled more gracefully:
* log the situation
* move to next task
* don't contact sentry.

PR #196, opened 2018-04-24
Fix failure after PR is forced pushed

slaykovsky:
When PR is forced pushed when test is running. The test will be
finished and then PR-CI will error out on status update step with

ERROR: __main__: Can't parse status data.

It is expected as the force push removed the status. But it
causes upload of data to sentry.io.

This patch adds EnvironmentError exception catching into execute
method of a Task class and raises ReferenceError when commit status
is not found. The error logs at a warning level, then runner's
execution proceeds.

Fixes https://github.com/freeipa/freeipa-pr-ci/issues/194

### DIFF
--- a/github/internals/entities.py
+++ b/github/internals/entities.py
@@ -677,7 +677,15 @@ class Task(object):
 
         result = self.job(dependencies_results)
 
-        status = world.poll_status(self.pr_number, self.name)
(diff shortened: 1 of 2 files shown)


stlaz:
LGTM.

PR #198, opened 2018-05-07
Kill Vagrant processes and VMs with psutil

slaykovsky:
The cleanup patch which should fix the cleanup was apparently wrong.

This one removes pkill call and uses psutil module instead to do the
cleanup.

### DIFF
--- a/tasks/common.py
+++ b/tasks/common.py
@@ -1,18 +1,24 @@
 import abc
 import collections
 import errno
-import jinja2
 import logging
 import os
-import psutil
+import signal
 import subprocess
 import threading
+from typing import Callable, List, Text
+
+import jinja2
+import psutil
 
 from . import constants
 
-
 LOG_FILE_HANDLER = None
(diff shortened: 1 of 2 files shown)


tasks/common.py
```diff
@@ -214,3 +220,50 @@ def create_file_from_template(template_path, dest, data):
...
+def kill_vagrant_processes() -> None:
+    """Kills all of Vagrant processes"""
+    return kill_processes(get_ruby_processes(), lambda f: "vagrant" in f.path)
+
+
+def kill_vagrant_vms() -> None:
+    """Kills all of Vagrant created VMs"""
+    def predicate(f: PopenFileType) -> bool:
+        return any([
+            n in f.path for n in ["master", "replica", "controller", "client"]
```
Tiboris: Here too or maybe a comment.

felipevolpone:
The code also LGTM, I'd change only the short variable names, as @Tiboris proposed.
Once @Tiboris ack it, we are ready to merge it :)

PR #200, opened 2018-05-14
WIP: ipa-run-webui-tests: changes to accept regexes in test definition

Rezney:
Currently when setting regexes in test definition in prci yaml
those are not properly expanded after coming from vagrant ssh
command and through ipa-run-webui-tests wrapper. Thus we are
taking different approach and getting them from tests directory
rather then passing it directly to pytest.

### DIFF
--- a/tasks/tasks.py
+++ b/tasks/tasks.py
@@ -319,10 +319,11 @@ class RunWebuiTests(RunPytest):
     def execute_tests(self):
         self.execute_subtask(
             PopenTask(['vagrant', 'ssh', '-c', (
-                'ipa-run-webui-tests {test_suite} '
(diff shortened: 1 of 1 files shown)


Rezney:
@felipevolpone could you please check this patch when you find some free cycles? Thanks a lot...

felipevolpone:
After fixing the `export IPATESTS_PATH=` thing, I put the test to run again. This is what I got: 
https://fedorapeople.org/groups/freeipa/prci/jobs/d8a2d8b2-5875-11e8-b7c4-fa163e8ca376/runner.log.gz

To be 100% honest, it's not clear to me what's happening, but it seems that it failed to find the tests to run.

PR #202, opened 2018-05-29
Define `with_python2 1` for rpm building

stlaz:
FreeIPA will not build python2-* rpms by default soon but we
still may want to test those that are left.

-----------

This should hopefully allow for https://github.com/freeipa/freeipa/pull/1563 to proceed and should not affect any other builds.


stlaz:
Obsoleted by https://github.com/freeipa/freeipa/pull/1976 which adds `with_python2 1` by default on Fedora. We'll get rid of Python 2 in a single step later.

PR #204, opened 2018-05-31
Reverting firewalld configuration

slaykovsky:
As the Fedora 28 doesn't have Python 2 bindings to firewalld and
ansible needs them by default to manage firewalld I'm removing
the firewalld configuration from the deployment process of the
runner.

Please note that as the part of FreeIPA team you should not worry
about it as our infrastructure has external firewall which manages
security rules.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,15 @@ from the up-to-date code should fix such issues.
 
 ### Side notes about deployment
 
+#### Firewall configuration
+If it happened you use firewalld on your machine you wish to install runner
+on you should open ports for nfs, ipp, rpc-bind and mountd. Otherwise Vagrant
+will not mount the shared volume.
+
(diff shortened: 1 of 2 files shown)


tomaskrizek:
I don't think it's wise to remove the firewall configuration.

Use of `firewalld` ansible module could be omitted in favor of using `shell` module with `firewall-cmd` command, which doesn't require python2 bindings.

slaykovsky:
> I don't think it's wise to remove the firewall configuration.

As I wrote in the commit message, we already have an external firewall that controls security groups.
So it's useless to have another one running on hosts.

tomaskrizek:
I'm not sure if the priorities of the project changed, but in the past, we've also considered the use case where users/developers can deploy the runner themselves. In those cases, automated firewall setup would be desirable.

Rezney:
@pvoborni can we merge this one? @slaykovsky could you please rebase?

PR #205, opened 2018-06-06
Changing volume caching from 'none' to 'unsafe'

slaykovsky:
Qcow2 file backend can be significantly slow with cache mode 'none'
setting. This happens because the actual file data is fragmented. In
other words, the qcow2 file remain a sparse one with only short stroke
of allocation (for metadata).

So it's better to have an aggressive VFS RAM caching to have a better
random access speed.

The key aspect of this unsafe mode, is that all flush commands from the
guests are ignored. Using this mode implies that the user has accepted
the trade-off of performance over risk of data loss in the event of a
host failure. As we are using virtual machines only for a short period
of time it's safe to use it.

### DIFF
--- a/templates/vagrantfiles/Vagrantfile.build
+++ b/templates/vagrantfiles/Vagrantfile.build
@@ -21,8 +21,7 @@ Vagrant.configure(2) do |config|
         # Disable graphics
         domain.graphics_type = "none"
 
-        # Recommended for remote NFS storage
-        domain.volume_cache = "none"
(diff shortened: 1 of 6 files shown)


slaykovsky:
Hello @rcritten 

> The tests that spawned the VMs will need to be re-executed anyway, right?

Yes, exactly.

> I think unsafe is fine

Cool, thanks.

rcritten:
There is no ack label so I'll just comment: ack

PR #208, opened 2018-06-13
remote_storage: switch from Fedora infra to AWS S3

Rezney:
Currently we are updating test results in Fedora infra where the results are
    directly served as part of web server directory listing capabilities. This is
    not case of AWS S3.
    In order to simulate same environment we are generating "index.html" using
    Jinja2 template and putting it into every directory. Then we upload particular
    job directory using awscli (which does parallelism for us) under S3 bucket
    "jobs" prefix (directory). Currently 2 awscli commands are needed in order to
    set correct encoding for gzip files. Content types are set automatically using
    "/etc/mime.types" file.
    At the end we create root jobs index to list all jobs in the bucket.
There is also improvement that we can see PR#, author, task name and result together with UUID and filter using datatables.

### DIFF
--- a/tasks/remote_storage.py
+++ b/tasks/remote_storage.py
@@ -1,9 +1,220 @@
 import os
 import re
+import concurrent
+import socket
+from concurrent.futures import ThreadPoolExecutor
+from datetime import datetime
 
-from .common import PopenTask, TaskException
-from .constants import (FEDORAPEOPLE_KEY_PATH, FEDORAPEOPLE_DIR, UUID_RE,
-                        JOBS_DIR)
+import logging
+import boto3
+from botocore.exceptions import ClientError
+from jinja2 import Template
+
+from .common import PopenTask, TaskException, FallibleTask
+from .constants import (CLOUD_JOBS_DIR, CLOUD_JOBS_URL, CLOUD_URL, CLOUD_DIR,
+                        CLOUD_BUCKET, UUID_RE, JOBS_DIR, TASKS_DIR)
+
+
+"""
+Previously we were updating test results in Fedora infra where the results were
+directly served as part of web server directory listing capabilities. This is
+not case of AWS S3.
+In order to simulate same environment we are generating "index.html" using
+Jinja2 template and putting it into every directory. Then we upload particular
+job directory using awscli (which does parallelism for us) under S3 bucket
+"jobs" prefix (directory). Currently 2 awscli commands are needed in order to
+set correct encoding for gzip files. Content types are set automatically using
+"/etc/mime.types" file.
+At the end we create root jobs index to list all "freeipa" repo PR jobs in the
+bucket.
+"""
+
+
+def get_jobdir_metadata(job_uuid):
+    """
+    Job metadata for root index.
+    """
+    try:
+        client = boto3.client('s3')
+        job = client.head_object(Bucket=CLOUD_BUCKET, Key=job_uuid)
+        name = job_uuid.split(os.sep)[-2]
+        if re.match(UUID_RE, name):
+            mtime = job.get('LastModified')
+            metadata = job.get('Metadata', dict())
+            repo_owner = metadata.get('repo_owner')
+            pr_number = metadata.get('pr_number')
+            pr_author = metadata.get('pr_author')
+            task_name = metadata.get('task_name')
+            returncode = metadata.get('returncode')
+            if repo_owner == 'freeipa':
+                return {'name': name,
+                        'mtime': mtime,
+                        'pr_number': pr_number,
+                        'pr_author': pr_author,
+                        'task_name': task_name,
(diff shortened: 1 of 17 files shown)


Rezney:
@slaykovsky Please take a look :) Thanks...

slaykovsky:
Hello @Rezney 

As it was discussed offline, the patch LGTM. You get my strong 👍 on this one as I saw its magic. 

However, there are some linting/formatting problems that should be fixed.

You can set up a linting pre-commit hook to make your life much easier. You can take some of [those](https://pre-commit.com/hooks.html), for example. In the addition you can read a very nice [PEP 8](https://www.python.org/dev/peps/pep-0008/) to get the _instinct_ on how to write readable Python code 😉

So I want to add a link on a PR where you tested these changes which contains links on the AWS so @pvoborni @stlaz @flo-renaud  and @Tiboris will see it and be thrilled by it as I was. This will also make a review/merging process faster 😊 >> https://github.com/Rezney/freeipa/pull/12 <<

Rezney:
@slaykovsky thanks for the review. Your comments should be hopefully addressed...

Rezney:
@slaykovsky Thanks for the review. I had to do some changes and refactoring due to boto3 slowness (when not parallel). E.g. switched back to using "awscli" for jobdir uploading (solved the issue with "directory" not created with `assign_jobdir_metadata()`), divided CloudUpload and CreateRootIndex for better handling purposes, using Futures (as discussed) for getting "directory" metadata.

Please let me know if this is ok. Especially if exception handling and logging is fine or it requires any improvement.

Cheers!

tomaskrizek:
It looks like the `*.gz` files from aws s3 can't be displayed directly in the browser, which can be convenient. I'm not sure if you can modify the site/server that serves the files, but for reference, this is how it was solved in the fedora infrastructure: https://pagure.io/fedora-infrastructure/issue/6074

Rezney:
@tomaskrizek Thanks for the pointer to the ticket but I am afraid that it will not help in this case cause I presume they fixed this on apache side. I was setting right content-type and encoding for "hostname.gz" in AWS S3 console manually but it is still not working. Previously I was testing everything with "runner.log.gz" so did not realize just plain "*.gz" is not working.

However as most important "runner.log.gz" and "build.log.gz" (also together with images from WebUI tests) are working I would not consider this a blocker.

@pvoborni what do you think?

Rezney:
@slaykovsky @tomaskrizek @pvoborni - "workaround-ed" the issue by adding a hostname directly to index. So there is now no need to open the file.

http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/4ab300ac-782f-11e8-94e6-fa163ee86de1/

Currently this patch together with @slaykovsky `s PR#206 ("qemu" killing) is running in production in order to properly test it.

@flo-renaud  ^^^ FYI

tasks/remote_storage.py
```diff
@@ -1,9 +1,221 @@
...
+        job = client.head_object(Bucket=CLOUD_BUCKET, Key=job_uuid)
+        name = job_uuid.split(os.sep)[-2]
+        if re.match(UUID_RE, name):
+            mtime = job.get('LastModified', None)
+            metadata = job.get('Metadata', dict())
+            repo_owner = metadata.get('repo_owner', None)
+            pr_number = metadata.get('pr_number', None)
+            pr_author = metadata.get('pr_author', None)
+            task_name = metadata.get('task_name', None)
+            returncode = metadata.get('returncode', None)
```
slaykovsky: Hello, dictionary get method will give you `None` anyway :) https://docs.python.org/3/library/stdtypes.html#dict.get

Rezney: Yep, good eye...fixed

Rezney:
Also added "yml yaml" to "mime.types" so we can open also "yaml" files as text. The "gz" files are covered by awscli. So currently only extension-less files are not being served as plain-text which can be solved later. This change has been running in production for 3 days and seems to be working fine.

http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/

@flo-renaud @slaykovsky is it ok?

tasks/remote_storage.py
```diff
@@ -1,9 +1,221 @@
...
+    try:
+        client = boto3.client('s3')
+        objects = []
+        with ThreadPoolExecutor(max_workers=100) as executor:
+            todo = []
+            for job in get_jobs():
+                future = executor.submit(get_jobdir_metadata, job)
+                todo.append(future)
+            for future in concurrent.futures.as_completed(todo):
+                res = future.result()
```
slaykovsky: Here you can check iff `res` is not `None` to append it to objects, with that you will get rid of line 101

slaykovsky:
LGTM & works well. 👌🏻

PR #209, opened 2018-06-19
Adding netoarmando to whitelist

rcritten:


netoarmando:
@pvoborni Could you merge it? Is there any other step to do after this?

PR #210, opened 2018-06-21
open/close PR tool: install crontabs on the host

flo-renaud:
The Ansible role automation/openclose_pr is used to configure a
machine responsible for creating nightly PRs.
This role configures a cron job launching the script
openclose_pr.py every night at 00:23. The role currently does not
install crontabs package on the machine, which leads to deployment
failure.

This commit modifies the Ansible role so that it also requires
crontabs package installation.


Rezney:
LGTM

PR #201, opened 2018-05-23
Making PR CI nightly tests definition editable by FreeIPA contributors

felipevolpone:
Now the .freeipa-pr-ci.yaml it's just a symlink to a file that actually has the definition of the tests. This commit changes the tool that schedules the nightly tests.

The major change is that the "--prci_config" parameter now receives a relative path that will change to where the .freeipa-pr-ci.yaml link points

PR #211, opened 2018-06-22
Update the documentation for PR scheduler tool

flo-renaud:
Add a separate documentation for PR scheduler tool, explaining the python script and the ansible playbook used to deploy the tool.


doc/prscheduler.md
```diff
@@ -0,0 +1,155 @@
...
+containing repo info and credentials
+  * installation of the python script in `/root/openclose_pr/openclose_pr.py`
+  * creation of the test definition file in
+`/root/openclose_pr/.freeipa-pr-ci.yaml`
+  * preparation of ssh configuration in order to access github.com
+  * creation of a clone repo from $repo_owner/freeipa.git to
+`/root/openclose_pr/freeipa/`
+  * configuration of the cron job to run daily at 23:00
+
+#### Current scheduling
```
Rezney: I would again mention here that we are usually opening nightly PRs from "freeipa-pr-ci" GitHub user.

Rezney:
Otherwise looks good...

doc/prscheduler.md
flo-renaud: Good point, updated.

flo-renaud:
@Rezney I updated the PR with your proposal, thanks!

pvoborni:
Nice addition. Looks good to me. But I did not thoroughly check all the args.  As haven't run the scripts myself so far. 

It will need to be updated after #201 or as part of #201 - depends on what is merged first.

flo-renaud:
Doc updated to reflect the changes introduced by #201 .

doc/prscheduler.md
```diff
@@ -0,0 +1,157 @@
...
+`/root/openclose_pr/.freeipa-pr-ci.yaml` to $prci_config
+  * preparation of ssh configuration in order to access github.com
+  * creation of a clone repo from $repo_owner/freeipa.git to
+`/root/openclose_pr/freeipa/`
+  * configuration of the cron job to run daily at 23:00
+
+#### Current scheduling
+
+Currently we are using a machine called prci-automation in our OpenStack lab,
+with cron job starting at 23:45 every monday, wednesday and friday,
```
Rezney: I think we should decide on what to hard-code here. Either everyday or mo, we, fri. 3x per week is probably ok.

flo-renaud: I agree it would be better to align the code in freeipa-pr-ci and the lab config. We should use mon, we, fri in the lab, meaning that I need to remove that comment and modify the crontab in the  task /ansible/roles/automation/openclose_pr/tasks/main.yml to be consistent with this decision.

flo-renaud:
Updated to reflect the change proposed in PR #222 Scheduler for nightly tests: run on Mon, Wed and Fri

Rezney:
Looks good to me now. @pvoborni are the recent changes ok with you?

pvoborni:
Just noticed this is pending on me. If you think it is fine then it is fine and can be pushed. Small issues can be always fixed later.

PR #213, opened 2018-07-04
machine_template: switch to paramiko system package

Rezney:
Due to gssapi issue https://github.com/paramiko/paramiko/issues/1069
in paramiko we were using fixed paramiko==2.2.1 installed by pip. As
the fix for the issue above was backported to Fedora we start using
the recent F28 system package.


flo-renaud:
LGTM.

PR #214, opened 2018-07-04
run "vagrant halt" during template create with privilege escalation

Rezney:
At the end of creating new vagrant box template script
"create-template-box" fails with permissison denied as we do
not have privileges to halt vagrant machine. Adding "become" also
to "shutdown template VM" task solves the issue.

### DIFF
--- a/ansible/roles/box/package/tasks/main.yml
+++ b/ansible/roles/box/package/tasks/main.yml
@@ -10,6 +10,7 @@
 # Turn template VM off and on again. Created vagrant box may not boot if 
 # this step is omitted.
 - name: shutdown template VM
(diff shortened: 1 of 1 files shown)


slaykovsky:
I'm **just** wondering why escalation is even needed there. I mean `template_box_dir` is in `/tmp` anyway and Vagrant has no problem with working under the non-privileged user.

So I can purpose to eliminate privilege escalation so in the future, you can use unmodified Fedora image on OpenStack and just add `fedora` user to a `libvirt` and `qemu` groups in Heap.

But anyway, this one fixes the issue and I'm okay with that.

PR #216, opened 2018-07-18
Fix editable PRCI nightly definition change

Rezney:
This patch fixes 8a2a771f4 where we were wrongly changing "world"
and thus second iteration of PR yielded instead of content of
symlink content of the yaml file and thus failing.


Rezney:
@slaykovsky thanks for helping with hunting this one. @tiran could you please take a look?

PR #217, opened 2018-07-26
root_index: switch from object metadata to AWS DynamoDB

Rezney:
We are using jobs root index for a convenience of easily
finding and filtering jobs results, however getting data from
object metadata is very slow and time consuming which can result
in task being timed-out. DynamoDB as part of Amazon AWS stack seems
to be more suitable for this task.

As this functionality is not part of main PRCI job a run should not
fail on this.

### DIFF
--- a/tasks/tasks.py
+++ b/tasks/tasks.py
@@ -93,7 +93,10 @@ class JobTask(FallibleTask):
         self.compress_logs()
         if self.publish_artifacts:
             self.upload_artifacts()
-            self.create_root_index()
+            # list only "freeipa" repo PRs in root index
+            # (Jobs of PRs against forks are not listed)
+            if self.repo_owner == 'freeipa':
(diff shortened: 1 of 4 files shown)


tasks/tasks.py
```diff
@@ -119,11 +122,20 @@ def create_root_index(self):
...
-            self.execute_subtask(CreateRootIndex(self.repo_owner))
+            self.execute_subtask(
+                CreateRootIndex(uuid=self.uuid,
+                                repo_owner=self.repo_owner,
+                                pr_number=self.pr_number,
+                                pr_author=self.pr_author,
+                                task_name=self.task_name,
+                                returncode=self.returncode,
+                                timeout=30,
+                                raise_on_err=False))
```
slaykovsky: The last one is to eliminate `raise_on_err`

PR #219, opened 2018-07-30
Add random number generator passthrough

Rezney:
Configure vagrant to use /dev/random passthrough and then configure
haveged on the runners to speed up entropy.


Rezney:
@flo-renaud  thanks for reviewing...

flo-renaud:
From a syntax standpoint, the fix is OK and the tests run properly.
Although I cannot certify that this modification fixes the issue of long installation times, I think it's worth giving it a try.

Rezney:
@flo-renaud what shall we do about this one? Close?

PR #222, opened 2018-08-16
Scheduler for nightly tests: run on Mon, Wed and Fri

flo-renaud:
The current configuration launches the nightly tests everyday. With this
fix, the nightly tests will be run only three times a week, on Monday,
Wednesday and Friday.


Rezney:
LGTM

PR #223, opened 2018-08-31
add f28 ansible var

Rezney:
Add missing f28 ansible var which was so far omitted.


flo-renaud:
Seems good to me. Maybe you could also leverage this PR to create ansible/vars/ipa_branches/ipa-4-7.yml as it has not been created yet, either.

Rezney:
@flo-renaud should be ok now. Sorry for the delay...

flo-renaud:
LGTM, ACK.

PR #225, opened 2018-09-25
Remove ipa python2 packages separately

Rezney:
As in the current master we stopped building python2 packages,
any attempt to uninstall them explicitly will fail. In this
case we differentiate uninstallation per branch.

### DIFF
--- a/ansible/roles/machine/setup/tasks/install_packages.yml
+++ b/ansible/roles/machine/setup/tasks/install_packages.yml
@@ -19,14 +19,20 @@
     - freeipa-server-common
     - freeipa-server-dns
     - freeipa-server-trust-ad
-    - python2-ipaclient
-    - python2-ipalib
-    - python2-ipaserver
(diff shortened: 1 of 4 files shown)


ansible/roles/machine/setup/tasks/install_packages.yml
```diff
@@ -19,15 +19,21 @@
...
 
+# this can be removed when python2 is gone completly
+- name: remove freeipa python2 packages seperately
+  shell: "rpm -e --nodeps '{{ item }}'"
+  with_items:
+    - python2-ipaclient
+    - python2-ipalib
+    - python2-ipaserver
+    - python2-ipatests
+  when: with_python2 is defined
```
flo-renaud: To be completely safe, I would replace with:
`when: with_python2 is defined and with_python2 == 1`

Rezney:
@flo-renaud  thanks for the review. Fixed...

PR #226, opened 2018-09-27
Vagrant automation

dhnunes:
Hello People,

This PR implements a simple method that creates the topology definition file in a dynamic way.

To do this, first, we need to implement a new topology definition at trigger file from freeipa:

topologies:
  build: &build
    build:
      cpus: 2
      memory: 6000

    master0:
      cpus: 4
      memory: 6700
    replica0:
      cpus: 2
      memory: 3000
    client0:
      cpus: 2
      memory: 1024

With this format and the template implemented on e63aa51, the function will append each machine declared in topology ( master0, replica0, etc ), to Vagrantfile.topology_name, that will be later used to per job execution.

### DIFF
--- a/github/internals/entities.py
+++ b/github/internals/entities.py
@@ -6,6 +6,7 @@ from enum import Enum, unique
 from random import randint
 from time import sleep
 from typing import Callable, ByteString, Dict, List, Text, Tuple, SupportsFloat
+import os
 
 import psutil
 import pytz
@@ -37,6 +38,40 @@ SENTRY_URL = (
 # until the reset time will come.
 EPHEMERAL_LIMIT = 60
 STALE_TASK_EXTRA_TIME = 60
+
+VAGRANT_FILES_DEFAULT_DIR = "/root/freeipa-pr-ci/templates/vagrantfiles/"
+VAGRANT_TEMPLATE_FILE = "Vagrantfile.template"
+
+
+def _vagrant_definition_creator(world):
+
+    with open(VAGRANT_FILES_DEFAULT_DIR + VAGRANT_TEMPLATE_FILE) as file:
+        text = file.read()
+
+        for topology in world.keys():
+
+            vagrantfile = open(VAGRANT_FILES_DEFAULT_DIR +
+                               "Vagrantfile.{}".format(topology), "w")
+
+            vagrantfile.write(text)
+
+            for host in world[topology].keys():
+
+                vagrantfile.write('    config.vm.define "{}"  do ',
+                                  '|{}|\n'.format(host, host))
+                vagrantfile.write('      {}.vm.provider :libvirt do',
+                                  '|domain|\n'.format(host))
+
+                for inner, configuration in world[topology][host].items():
+
+                    vagrantfile.write('          domain.{} = ',
+                                      '{}\n'.format(inner, configuration))
+
+                vagrantfile.write('      end\n')
+                vagrantfile.write('    end\n')
+
+            vagrantfile.write("end")
+            vagrantfile.close()
 
 
 def sentry_report_exception(context: Dict):
@@ -477,6 +512,8 @@ class PullRequest(object):
         task_link = self.__get_tasks_file_content(world)
         self.tasks_path = task_link.decode()
         tasks_file_content = self.__get_tasks_file_content(world)
+
+        _vagrant_definition_creator(yaml.load(tasks_file_content)["topologies"])
 
         try:
             return yaml.load(tasks_file_content)["jobs"]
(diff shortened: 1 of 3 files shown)


Rezney:
@dhnunes Hi, thanks for the patch. Is there any reason for playing with permissions? 

E.g. "prci.py" lost execution permission and while testing the patch I am getting:

`[root@prci ~]# journalctl -f -u prci
-- Logs begin at Mon 2018-10-01 15:37:37 CEST. --
Oct 01 17:17:30 prci.ipa.test bash[64403]: /bin/bash: /root/freeipa-pr-ci/github/prci.py: Permission denied
<snip>
`

`
[root@prci ~]# ll freeipa-pr-ci/github/
total 28
-rw-r--r--. 1 root root    0 Oct  1 16:26 __init__.py
drwxr-xr-x. 3 root root 4096 Oct  1 16:26 internals
-rw-r--r--. 1 root root 6686 Oct  1 16:26 open_close_pr.py
-rw-r--r--. 1 root root 9447 Oct  1 16:26 prci.py
drwxr-xr-x. 2 root root 4096 Oct  1 16:26 tests
`

Please also fix the yaml example in the description as the formatting/indentation got broken.

@f-trivino and @pvoborni will take a look from code point of view.

Cheers!

dhnunes:
@Rezney Hello! Thanks for the feedback.

There is no reason for change the permissions, I believe I did some confusion with this changes when submitting the PR. I will revert this changes.

Here is the example yaml with correct Indentation ( Forget backticks, me-- ):

```yaml
topologies:
  build: &build
    build:
      cpus: 2
      memory: 6000

    master0:
      cpus: 4
      memory: 6700
    replica0:
      cpus: 2
      memory: 3000
    client0:
      cpus: 2
      memory: 1024
```

dhnunes:
@Rezney I've rebased and removed files sent by mistake when PR was open.

@f-trivino  , @pvoborni , Fell free to ask anything related to changes, feedback is highly appreciated!

Thanks People

Rezney:
`
Oct 02 09:59:50 prci.ipa.test bash[74052]: ERROR: __main__: write() takes exactly one argument (2 given)
`

The culprit seems to be writing to Vagrantfile:

`
vagrantfile.write('    config.vm.define "{}"  do ',
                                  '|{}|\n'.format(host, host))
`

My advice here would be using Jinja templating language instead. You can find example here:

https://github.com/freeipa/freeipa-pr-ci/blob/18d0b00bb79b0d354cb72b6a71a64524e21f2958/tasks/remote_storage.py#L53

Documentation: http://jinja.pocoo.org/docs/2.10/

dhnunes:
@Rezney  ack, thanks for the update.

About the error:

```
vagrantfile.write(' config.vm.define "{}" do ', '|{}|\n'.format(host, host))
```

There is missing double quotes around the second curly brackets, this is why you have faced this error.

BTW, your mention for using the Jinja template sounds good, I will take a look in how I can implement this way.

Thanks ! :+1:

pvoborni:
Few additional comments: WRT commit messages - good resources how to write good ones are:
* https://chris.beams.io/posts/git-commit/
* http://superuser.openstack.org/articles/splitting-up-git-commits/
* https://wiki.openstack.org/wiki/GitCommitMessages

On the behavior side, apart from other comments: In the original topology definition, cpu and memory values were used to check if a runner has enough resources to run a task with such topology (there can be different sizes of PR-CI runners) - it was a sum for all vms, see https://github.com/freeipa/freeipa-pr-ci/blob/master/github/internals/entities.py#L279 . There is no behavior change regarding this check in this PR thus it will not work correctly. With the new behavior, when memory and cpu is defined per host, the code needs to be adjusted, e.g. to sum the values. If the value is not provided then a default should be added and also a value for controller needs to be added to the sum.

dhnunes:
@pvoborni,

I was able to identify the point you had mention about the resources usage from all vms. Before my changes, the second line in the following log was 0 CPU and 0 Memory, after the change, it takes apropriate format:

```
INFO: __main__: fedora-28/build PR#4 is successfully locked.
INFO: __main__: Available resources: 2 CPU, 1237.93359375MB
INFO: root: Initializing job 1ff63aca-d7cd-11e8-9eb8-00155d770102
```

That undesired behavior was happening because the dict with new format of yaml was ben correctly passed to Topology Factory, but since I've introduced a new level in yaml, the new dict is comming in a different format:

```
{'build': {'name': 'build', 'cpus': 2, 'memory': 6000}}
```

The code used to parse this dict inside the factory, looks like this one:

```
name=dict_data.get("name"),
memory=dict_data.get("memory"),
cpu=dict_data.get("cpu")
```

But, since the dict now has a new format, where:

- Topology name is the key;
- And Topology values are a new dict of this key;

The old implementation won't work, since we have a new level.  After the changes, it's now working. Code is WIP, since is done, I will push.

@f-trivino ,

As recommended I've changed the declaration of default path using os module and moved to appropriate place in contants module.

Also, I'm working in a desing to implement it using jinja2 templates, since is done, I will push.

netoarmando:
Closing this due to lack of activity.

Issue #227, opened 2018-09-28
Harden prci when test def contains a wrong topology


flo-renaud:
When a PR defines a wrong topology in a test definition, (for instance `topology: master_1repl_1client` instead of `topology: *master_1repl_1client`), PRCI crashes with the following trace:
```
Traceback (most recent call last):
   File "/root/freeipa-pr-ci/github/prci.py", line 307, in <module>
     main()
   File "/root/freeipa-pr-ci/github/prci.py", line 282, in main
     for task in process_pull_request(world, pull_request, repo_url):
   File "/root/freeipa-pr-ci/github/prci.py", line 122, in process_pull_request
     pull_request.author, repository_url, task_data, JobDispatcher
   File "/root/freeipa-pr-ci/github/internals/entities.py", line 575, in __init__
     self.topology = Topology.from_dict(topology_data)
   File "/root/freeipa-pr-ci/github/internals/entities.py", line 251, in from_dict
     name=dict_data.get("name"),
 AttributeError: 'str' object has no attribute 'get'
```

PRCI should be more robust. It could even set the 'postponed' label to the pull request so that the PR is not picked up for processing by other PR-CI runners, and add a message to the PR explaining the issue.

flo-renaud:
Note: `topology: master_1repl_1client` is syntactically correct but prci is expecting a reference to a repeated node that will contain a dict.

PR #228, opened 2018-10-01
Harden PRCI against wrong test definitions

flo-renaud:
Fixes Issue #227 

### Do not crash when test definition is missing a *

The test definition should contain a dict for the topology definition. Usually we are using a repeated node,for instance:
`topology: *master_1repl`
If the test definition is missing the *, then the topology contains a string instead of a dict and PRCI crashes.
The fix properly handles this error.

### Update the test status when the test could not be run

When a test could not be run (wrong YAML content for instance), PRCI simply logs a message in the journal but the dashboard is not updated to reflect the issue.
This commit updates the task status to ERROR and prints "Test not executed: Wrong job definition " so that it does not get unnoticed.

Example:
update_prci_status

### DIFF
--- a/github/internals/entities.py
+++ b/github/internals/entities.py
@@ -251,6 +251,8 @@ class Topology(object):
     @staticmethod
     def from_dict(dict_data: Dict) -> "Topology":
         """Factory for Topology"""
+        if not isinstance(dict_data, dict):
+            raise JobYAMLError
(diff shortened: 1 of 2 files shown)


Rezney:
@flo-renaud thank you very much for the patch. Seems to be working fine. LGTM.

Issue #230, opened 2018-10-24
Log the hostname when PRCI fails to publish artifacts


flo-renaud:
When a PRCI runner fails to publish the logs, the status on the PR page simply states:
```
fedora-28/test_client_uninstallation — RunPytest Failed to publish artifacts
```

but there is no information about the machine where the runner failed. It could be nice to simply add the hostname, so that we are able to check the machine and repair the issue:
```
fedora-28/test_client_uninstallation — RunPytest Failed to publish artifacts from ci-large-0
```

flo-renaud:
Closing as the improvement has been implemented with commit https://github.com/freeipa/freeipa-pr-ci/commit/39a4381fe66f9c952c362696f392e9f52780ebc2

PR #231, opened 2018-10-24
PRCI: when runner fails to publish logs, print hostname

f-trivino:
https://github.com/freeipa/freeipa-pr-ci/issues/230

When a PRCI runner fails to publish the logs, the status on the PR page
simply states: "Failed to publish artifacts", but there is no information
about the machine where the runner failed. This PR adds the hostname, so
that is easy to check the machine and repair the issue.

### DIFF
--- a/tasks/tasks.py
+++ b/tasks/tasks.py
@@ -110,7 +110,9 @@ class JobTask(FallibleTask):
                             timeout=5*60))
         except Exception as exc:
             logging.debug(exc, exc_info=True)
-            raise TaskException(self, "Failed to publish artifacts")
+            hostname = socket.gethostname().split('.')[0] # don't leak fqdn
+            msg = 'Failed to publish artifacts from {}'.format(hostname)
+            raise TaskException(self, msg)
         else:
             self.remote_url = urllib.parse.urljoin(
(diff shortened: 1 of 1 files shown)


tasks/tasks.py
```diff
@@ -110,7 +110,9 @@ def upload_artifacts(self):
                             timeout=5*60))
         except Exception as exc:
             logging.debug(exc, exc_info=True)
-            raise TaskException(self, "Failed to publish artifacts")
+            msg = 'Failed to publish artifacts from {hostname}'.format(
+                hostname=socket.gethostname())
```
netoarmando: This could leak the FQDN, as pointed by [some comments](https://github.com/freeipa/freeipa-pr-ci/search?q=socket.gethostname&unscoped_q=socket.gethostname) I've found in the codebase: https://github.com/freeipa/freeipa-pr-ci/blob/81e6d9fa91e6acb18eb721eabd4c79c39320257d/tasks/remote_storage.py#L223-L224

f-trivino: you right!, thanks Armando, Will amend the commit.

flo-renaud:
Hi @f-trivino 
thanks for the patch, looks good to me.

PR #235, opened 2018-11-13
autocleaner: prune /etc/exports when deleting old PRCI job folders

f-trivino:
Entries in /etc/exports are not removed on vagrant destroy, hence
/etc/exports must be accordingly cleaned-up when deleting PRCI
job folders in order to avoid nfs-server issues.


autocleaner.py
```diff
@@ -91,9 +91,39 @@ def load_yaml(yml_path):
...
+        export_folder = line.partition(' ')[0].strip('"')
+	if export_folder not in dir_name:
+            exports.write(line)
+        else:
+            entries_deleted.append(export_folder)
+
+    # Write it out and close it.
+    exports.truncate()
+    exports.close()
+    return entries_deleted
```
flo-renaud: The method will not remove the comments surrounding the removed line. For instance:
```
# VAGRANT-BEGIN: 0 a839dcb2-45bf-42fa-a341-fc98ce9ae18d
"/root/freeipa-pr-ci/jobs/0012c21a-9d3e-11e8-a1c8-fa163e311250" 192.168.121.165(rw,no_subtree_check,all_squash,anonuid=0,anongid=0,fsid=1790925634)
# VAGRANT-END: 0 a839dcb2-45bf-42fa-a341-fc98ce9ae18d
```
will be replaced with
```
# VAGRANT-BEGIN: 0 a839dcb2-45bf-42fa-a341-fc98ce9ae18d
# VAGRANT-END: 0 a839dcb2-45bf-42fa-a341-fc98ce9ae18d
```
Would it be possible to also remove the comments around removed files?

autocleaner.py
```diff
@@ -91,9 +91,39 @@ def load_yaml(yml_path):
...
+            return None
+    except:
+	pass
+
+    exports = open('/etc/exports', 'r+')
+    content = exports.readlines()
+    exports.seek(0)
+    for line in content:
+        export_folder = line.partition(' ')[0].strip('"')
+	if export_folder not in dir_name:
```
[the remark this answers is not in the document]
f-trivino: yes :) thanks.

autocleaner.py
f-trivino: It was my intention, so we can verify that the cleanup is working, but I think you right... we do not want to left garbage on /etc/exports file... will apply the change.

f-trivino:
dhnunes, thanks for your review. You right!, we must prune /etc/exports and re-sync entries before deleting so we can prevent other kinds of issues... Will submit a new patch.

autocleaner.py
```diff
@@ -91,20 +91,71 @@ def load_yaml(yml_path):
...
+	pass
+
+    exports = open('/etc/exports', 'r+')
+    content = exports.readlines()
+    exports.seek(0)
+    for line in content:
+        if "# VAGRANT-BEGIN" not in line:
+            exports.write(line)
+        else:
+            next_item = content.index(line) + 1
```
f-trivino: we could implement this in a more efficient way but I do not want to complicate the source code for a very simple feature.

dhnunes: Hey Triviño !
Looks ok now !

autocleaner.py
```diff
@@ -91,20 +91,71 @@ def load_yaml(yml_path):
...
+    hence /etc/exports must be accordingly cleaned-up before deleting
+    PRCI job folders in order to avoid nfs-server issues.
+    """
+    entries_deleted = []
+    try:
+        if not os.geteuid()==0:
+            print("\nUnable to prune /etc/exports. Must be run as root\n")
+            return None
+    except:
+	pass
```
flo-renaud: nitpick, but the indentation is different from above

PR #240, opened 2018-11-29
autocleaner: fix taberror and add emacs gitignore entry

f-trivino:
autocleaner: fix taberror and add emacs gitignore entry


flo-renaud:
@f-trivino 
thanks for the patch, LGTM.

PR #242, opened 2018-12-11
configure sunrpc to leave kadmin port 749/TCP open

tiran:
The default value of sunrpc.min_resvport is 665. An NFS mount can block
the kadmin service port and cause IPA installation to fail.

Signed-off-by: Christian Heimes <cheimes@redhat.com>


abbra:
ACK. Looks good to me.

PR #245, opened 2019-01-07
Rawhide fixes

tiran:
Various fixes for rawhide / F30 PR-CI images.

* python2-dnf is no longer available, use Python 3 for Ansible
* Fix download URL for rawhide image
* ~Don't use Python dependency generator~
* Remove conflicting packages
* Fix box uploader
* Add Fedora 30 target
* Disable SELinux temporarily (due to https://github.com/freeipa/freeipa-pr-ci/pull/245#issuecomment-488171501)

### DIFF
--- a/ansible/create_template_box.yml
+++ b/ansible/create_template_box.yml
@@ -18,20 +18,17 @@
   vars_files:
     - vars/fedora/{{ fedora_version }}.yml
     - vars/ipa_branches/{{ git_branch }}.yml
+  gather_facts: false
   pre_tasks:
+    - name: install Python 3 deps for ansible modules
+      raw: dnf install -y python3-dnf python3-libselinux
+      become: yes
+    - name: gather facts
+      setup:
     - set_fact:
         template_box_name: "{{ flow  }}-{{ git_branch }}-f{{ fedora_version }}"
   roles:
     - box/prepare
-
-# HACK: workaround to install Python2 that is still required by Ansible
-- hosts: image_box
-  gather_facts: false
-  tasks:
-    - name: install python and deps for ansible modules
-      raw: test -e /usr/bin/python || dnf install -y python2 python2-dnf libselinux-python
-    - name: gather facts
-      setup:
 
 - hosts: image_box
   name: modify vagrant box template
--- a/ansible/roles/machine/setup/tasks/install_packages.yml
+++ b/ansible/roles/machine/setup/tasks/install_packages.yml
@@ -4,6 +4,15 @@
     name: '*'
     state: latest
 
+# temporary workaround for rawhide issue
+- name: remove packages that cause conflicts on rawhide
+  dnf:
+    name:
+      - gdbm
+      - python2-chardet
+    state: absent
+  when: fedora_version == 'rawhide'
+
 - name: install freeipa rpms with dependencies
   shell: dnf install -y /vagrant/rpms/*.rpm
 
--- a/ansible/roles/machine/setup/tasks/install_packages.yml
+++ b/ansible/roles/machine/setup/tasks/install_packages.yml
@@ -49,12 +49,15 @@
     state: latest
   with_items:
     - PyYAML  # ipa-vagrant-ci dependency
+    - python3-pyyaml
     - vim
     - NetworkManager
     - xorg-x11-server-Xvfb
     - firefox
     - python3-paramiko
     - firewalld
+    - python3-dnf  # Ansible dependency
+    - python3-libselinux  # Ansible dependency
 
 # Configure firewalld (make sure that ssh is enabled)
 # This is using the command line tool till python2/3 firewalld import issues
--- a/ansible/roles/machine/setup/tasks/install_packages.yml
+++ b/ansible/roles/machine/setup/tasks/install_packages.yml
@@ -11,7 +11,7 @@
       - gdbm
       - python2-chardet
     state: absent
-  when: fedora_version == 'rawhide'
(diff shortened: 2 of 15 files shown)


ansible/create_template_box.yml
```diff
@@ -6,29 +6,25 @@
...
       prompt: "git branch"
       private: false
       default: "master"
   vars_files:
     - vars/fedora/{{ fedora_version }}.yml
     - vars/ipa_branches/{{ git_branch }}.yml
+  gather_facts: false
   pre_tasks:
+    - name: install Python 3 deps for ansible modules
+      raw: dnf install -y python3-dnf python3-libselinux
```
netoarmando: This is intended to be executed on developer's machine, so this must add `become: yes` parameter.

Also , it could use `dnf` instead of `raw`.

netoarmando:
I've opened a PR to test the image built after these changes here: https://github.com/freeipa-pr-ci2/freeipa/pull/10.

netoarmando:
> I've opened a PR to test the image built after these changes here: freeipa-pr-ci2/freeipa#10.

Build failed:
```bash
2019-04-01 19:50:03,048    DEBUG  TASK [builder/build : add autoreconf to spec file] *****************************
2019-04-01 19:50:04,031    DEBUG  fatal: [builder]: FAILED! => {"changed": false, "msg": "Aborting, target uses selinux but python bindings (libselinux-python) aren't installed!"}
2019-04-01 19:50:04,032    DEBUG  
2019-04-01 19:50:04,032    DEBUG  msg: Aborting, target uses selinux but python bindings (libselinux-python) aren't installed!
2019-04-01 19:50:04,033    DEBUG  	to retry, use: --limit @/root/freeipa-pr-ci/ansible/build.retry
2019-04-01 19:50:04,033    DEBUG  
```

@tiran, Should this be fixed by having `python3-libselinux` pre-installed on the machine?

ansible/roles/builder/build/tasks/create_spec.yml
```diff
@@ -11,6 +11,13 @@
     regexp: "^Version:"
     line: "Version:        {{ build_version }}"
 
+- name: disable Python dependency generator
```
tiran: This block is no longer necessary. Please remove it.

ansible/roles/machine/setup/tasks/install_packages.yml
```diff
@@ -4,6 +4,15 @@
...
     state: latest
 
+# temporary workaround for rawhide issue
+- name: remove packages that cause conflicts on rawhide
+  dnf:
+    name:
+      - gdbm
+      - python2-chardet
+    state: absent
+  when: fedora_version == 'rawhide'
```
tiran: Just rawhide or also Fedora 30?

tasks/ansible.py
```diff
@@ -20,9 +20,15 @@ def _run(self):
 class AnsiblePlaybook(PopenTask):
     def __init__(self, playbook=None, extra_vars=None,
                  verbosity=None, **kwargs):
+        if self.extra_vars is None:
```
netoarmando:
```python
AttributeError: 'AnsiblePlaybook' object has no attribute 'extra_vars'
```
I'm changing this line to:
```python
if extra_vars is None:
```
However, only this method is using this object's attribute. Actually, the same applies to `self.verbosity` and `self.playbook`.

tasks/ansible.py
```diff
@@ -20,9 +20,15 @@ def _run(self):
...
+            extra_vars = {}
+        # use Python 3 as default Python interpreter for Ansible
+        extra_vars.setdefault(
```
netoarmando: This works only for `build` tasks, they are the only one calling this class, for the rest I think the best is to add that setting to the files:
- `/templates/ad.vars.yml`
- `/templates/run_pytest.vars.yml`

However, I tested it and hit the following error related to selinux:
```
2019-04-30 21:38:05,364    DEBUG  fatal: [controller]: FAILED! => {"changed": false, "checksum": "295e3e77dde23ab410afa683ad7589e9070387c6", "cur_context": ["system_u", "object_r", "nfs_t", "s0"], "gid": 0, "group": "root", "input_was": ["system_u", "object_r", "default_t", "s0"], "mode": "0744", "msg": "invalid selinux context: [Errno 95] Operation not supported", "new_context": ["system_u", "object_r", "default_t", "s0"], "owner": "root", "path": "/vagrant/.ansible_tmpusb2b4bgipa-test-config.yaml", "secontext": "system_u:object_r:nfs_t:s0", "size": 607, "state": "file", "uid": 0}
2019-04-30 21:38:05,375    DEBUG  
2019-04-30 21:38:05,375    DEBUG  msg: invalid selinux context: [Errno 95] Operation not supported
```
[Complete log here](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/195edcce-6ba9-11e9-9757-5254002f0d7e/runner.log.gz).

ansible/roles/machine/setup/tasks/install_packages.yml
netoarmando: Apparently, this is not required any more.

netoarmando:
I've updated what have been discussed until this point to a branch on my fork. Changes could be compared here: https://github.com/freeipa/freeipa-pr-ci/compare/master...netoarmando:rawhide_fixes

There is a problem with selinux I've mentioned in a comment [above](https://github.com/freeipa/freeipa-pr-ci/pull/245#issuecomment-478725702).

PR #256, opened 2019-03-11
Stop prci systemd service when unable to publish artifacts

f-trivino:
PR-CI publish artifacts fails when the runner gets out of space, but the systemd service
keeps running. This causes the next scheduled jobs to fail with the same reason. With this
commit, prci systemd service is disabled to avoid causing more troubles.

### DIFF
--- a/tasks/tasks.py
+++ b/tasks/tasks.py
@@ -4,6 +4,7 @@ import shutil
 import socket
 import urllib
 import uuid
+import subprocess
 
 from .ansible import AnsiblePlaybook
 from .common import (FallibleTask, TaskException, PopenTask,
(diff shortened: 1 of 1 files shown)


f-trivino:
@netoarmando do you think we could connect this error with Sentry? then we will receive a notification of the runner is being deactivated.

netoarmando:
> @netoarmando do you think we could connect this error with Sentry? then we will receive a notification of the runner is being deactivated.

Sentry collects exceptions, so in this case it's already working, not to notify us that the service was stopped, but that the artifacts couldn't be published.

However, with after your changes at least one new exception might be raised: `subprocess.TimeoutExpired`, when it happens Sentry will collect and aggregate that instead.

f-trivino:
> > @netoarmando do you think we could connect this error with Sentry? then we will receive a notification of the runner is being deactivated.
> 
> Sentry collects exceptions, so in this case it's already working, not to notify us that the service was stopped, but that the artifacts couldn't be published.
> 
> However, with after your changes at least one new exception might be raised: `subprocess.TimeoutExpired`, when it happens Sentry will collect and aggregate that instead.

yes, that was my suspicion... "failed to publish artifacts" is already present in Sentry.. so with this commit, it will mean that the runners has been disabled.

netoarmando:
> yes, that was my suspicion... "failed to publish artifacts" is already present in Sentry.. so with this commit, it will mean that the runners has been disabled.

Yes, except when stopping the service fails too. Anyway, I think your changes fix the scenario you've described.

PR #276, opened 2019-05-31
use @pki/master copr for f30

frasertweedale:
FreeIPA master now requires Dogtag 10.7.  The @pki/10.6-nightly COPR
is not even built for f30, let alone with pki-10.7 packages.  Use
the @pki/master COPR on f30 to pull in the latest Dogtag builds.

(This if my first pr-ci PR... do images have to be explicitly rebuilt
after a change like this gets merged, or does it happen automagically?)

### DIFF
--- a/ansible/vars/fedora/30.yml
+++ b/ansible/vars/fedora/30.yml
@@ -6,4 +6,5 @@ repo_fedora_enabled: 1
 repo_updates_enabled: 1
 repo_updates_testing_enabled: 1
 repo_freeipa_copr_enabled: 1
-repo_pki_testing_enabled: 1
+repo_pki_testing_enabled: 0
+repo_pki_master_enabled: 1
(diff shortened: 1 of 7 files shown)


netoarmando:
> (This if my first pr-ci PR... do images have to be explicitly rebuilt after a change like this gets merged, or does it happen automagically?)

@frasertweedale They need to be rebuild, manually. I'm building a new one based on your changes before merging this. 

Is this true for `rawhide` as well?

ansible/vars/fedora/30.yml
```diff
@@ -6,4 +6,4 @@ repo_fedora_enabled: 1
 repo_updates_enabled: 1
 repo_updates_testing_enabled: 1
 repo_freeipa_copr_enabled: 1
-repo_pki_testing_enabled: 1
```
netoarmando: Do not remove this, set it to zero instead, otherwise ansible complains:
```
TASK [builder/prepare : set default mock config] ***********************************************************************
fatal: [192.168.121.120]: FAILED! => {"changed": false, "msg": "AnsibleUndefinedVariable: 'repo_pki_testing_enabled' is undefined"}
```
You also need to add `repo_pki_master_enabled: 0` to all other templates (except `rawhide`).

netoarmando:
I tested these changes here: https://github.com/netoarmando/freeipa/pull/15 

@frasertweedale Should I add that repo to rawhide as well?

frasertweedale:
@netoarmando yes, it probably makes sense to add to rawhide too.

PR #303, opened 2019-08-07
Delete folders for jobs older than 3 days

netoarmando:
Runners are hitting 'out-of-space' issue. Something between 2 and 7GB are stored weekly.


f-trivino:
LGTM. Looks good to me.

PR #321, opened 2019-11-05
Add fedora 31 box settings

netoarmando:
- Fix tasks failing with recent Fedora hosts
  - These changes are backwards compatible, but since Fedora 30 those changes
are required to continue to build images.
- Add Fedora 31 variables


ansible/vars/fedora/31.yml
```diff
@@ -0,0 +1,3 @@
+---
+fedora_releasever: 31
```
f-trivino: why not fedora_release_server?

netoarmando: Legacy only.

PR #335, opened 2020-01-07
Stabilize provisioning environments for AD-related tests

wladich:
This PR is targeting execution of test_winsyncmigrate test suite but also improves stability for other AD-related tests.


wladich:
Test PR execitiong test_winsyncmigrate 100 times: https://github.com/wladich/freeipa/pull/5

There are 2 failures:
1. `Not enough virtual memory or paging file quota is available to complete the specified operation.` during establishing trust.
2. `Connection refused` while trying to connect to LDAP server in Windows server.

While result is still not perfect it is much more stable than without proposed patches.

netoarmando:
These changes have been deployed to production runners.

PR #343, opened 2020-01-31
Enable execution of tests with multiple AD instances

wladich:
This PR provides vagrant template and ansible playbooks for setting up Windows domain with subdomain and a treedomain. It also improves stability of provisioning of VMs.

Tested with all AD-related tests from freeipa: https://github.com/wladich/freeipa/pull/11

Changes in this PR:
* move winrm options from Vagrantfile to ansible host vars
  This will help to reduce copy-pasting when creating new templates with
  Windows instances.
* Increase WinRM timeouts for Vagrant and ansible to reduce number of communication errors
* Use standard ansible task instead of ad-hoc script to setup AD root domain
* replace hardcoded AD hostname and domain name with constants
* Add vagrantfile templates and ansible playbooks for 3 AD instances
  Also changed: stop playbook execution when any play fails - it is
  generally useless to continue Windows domains setup when one of play fails.
* install openssh on Windows without chocolatey
  Chocolatey server has very strict rate limiting policy, it offten banns our ip address for period of one hour.
This commit also moves ssh setup to separate role
* retry if creation of directory fails
  Directory for storing rpms lists is created inside sshfs-mount.
  This operation fails randomly, presumably due to race condition, as
  all linux machines are creating the mount almost simultaneously.
*  refactor repeating of VMs provisoning:
      * move retries for `vagrant up` and `vagrant provision` to single
      place to make logic more clear
      * retry `vagrant provision` without destroying VMs - provisioning of
      Windows machines sometimes fails with timeout errors which can be
      recovered with successive execution of playbooks. At the same time
      ansible playbooks are written in idempotent style, so repeated execution
      does not cause any problems.
      * increase number of retries when provisioning Windows VMs
      * add  delay before starting provisioning for Windows machines - this
      helps to partially skip initial rush of Windows maintenance processes
      * remove redundand logging of errors (it is already done in FallibleTask)


netoarmando:
@wladich Thanks for your work here! I'm reviewing this ASAP.

ansible/roles/machine/provision/templates/hosts
```diff
@@ -5,6 +5,13 @@
...
+
+{% for host, domain in [
+  (ad_root_dc_hostname, ad_root_domain),
+  (ad_child_dc_hostname, ad_child_domain),
+  (ad_tree_dc_hostname, ad_tree_domain)
+] %}
+{% if host in groups['all'] %}
+      {{ hostvars[host]['ansible_ip_addresses']|ipv4|first }} {{ host }}.{{ domain }}
 {% endif %}
+{% endfor %}
```
netoarmando: No newline at end of file.

ansible/roles/windows/ipa-ad-data/tasks/main.yml
```diff
@@ -1,82 +1 @@
...
-      OtherAttributes: "@{'uidNumber'='10048'; 'gidNumber'='10047'; 'loginShell'='/bin/sh'; 'homeDirectory'='/home/ad.test/upnuser'; 'gecos'='UPN User'}"
-
-- name: set UPN User primary group
-  include_role:
-    name: windows/ipa-ad-data
-    tasks_from: primary_group.yml
-  vars:
-    user: upnuser
-    group: testgroup
+- include_tasks: "{{ inventory_hostname }}.yml"
```
netoarmando: No newline at end of file.

tasks/vagrant.py
```diff
@@ -31,23 +31,54 @@ def __setup_provision(task):
...
+
+    while True:
+        try:
+            task.execute_subtask(
+                VagrantUp(timeout=None))
+            break
+        except PopenException:
+            if vagrant_up_retries:
+                logging.info(
+                    "Retrying to bring the machine up, %s retries left" %
```
netoarmando: nitpick: Could you please pass the variables as arguments to the logging functions or use the `str.format()`.
```python
logging.info("Retrying to bring the machine up, %s retries left", vagrant_up_retries)
```

ansible/roles/windows/sshd/tasks/main.yml
```diff
@@ -0,0 +1,38 @@
+- name: Check if sshd service exists
```
netoarmando: Since this was moved out from `windows/common/tasks/main.yml`, what do you think about moving it to `windows/common/tasks/sshd.yml`?

wladich: I think it is better to keep it as a separate role, this way it will be easier to reuse it in other project as all constants are contained in its own defaults directory.

netoarmando: Agreed.

PR #344, opened 2020-02-03
Add win_domain_group

amore17:
Added group 'group@group'
needed for PR: https://github.com/freeipa/freeipa/pull/4191


wladich:
Note: when test is implemented, the wiki page needs to be updated: https://www.freeipa.org/page/V4/AD_configuration_for_testing

PR #348, opened 2020-02-11
vagrant catalog provisioning fixes and improvements 

netoarmando:
- Add missing vagrant catalog tasks
  - SELinux was blocking nginx to talk to gunicorn socket.
  - Gnunicorn service wasn't enabled nor started after `gunicon.socket` service was started.
- Reset environment variable to use Vagrant's default catalog
  - This can be used to easily rollback PR-CI to use Vagrant's catalog if `prci-automation` is not available.

### DIFF
--- a/ansible/roles/automation/box_hosting/tasks/main.yml
+++ b/ansible/roles/automation/box_hosting/tasks/main.yml
@@ -26,6 +26,11 @@
       - nginx
       - python3-gunicorn
 
+- name: set selinux to permissive
+  selinux:
+    policy: targeted
+    state: permissive
+
 - name: copy nginx settings
   copy:
     src: nginx.conf
@@ -66,7 +71,7 @@
   service:
     enabled: true
     state: started
-    name: gunicorn.socket
+    name: gunicorn
 
 - name: reload nginx service
   systemd:
--- a/ansible/roles/runner/tasks/custom_vagrant_catalog.yml
+++ b/ansible/roles/runner/tasks/custom_vagrant_catalog.yml
@@ -1,18 +1,27 @@
 ---
 # Configure host to fetch boxes from custom catalog (prci-automation)
-- name: Add prci-automation entry to /etc/hosts
-  lineinfile:
-    path: /etc/hosts
-    regexp: "prci-automation$"
-    line: "{{ custom_vagrant_catalog }} prci-automation"
-
-- name: Set "VAGRANT_SERVER_URL" Environment variable
+- name: Reset "VAGRANT_SERVER_URL" Environment variable
   ini_file:
     path: /etc/systemd/system/prci.service
     section: Service
     option: Environment
-    value: "VAGRANT_SERVER_URL='http://prci-automation'"
-    mode: 0700
+    state: absent
+
+- block:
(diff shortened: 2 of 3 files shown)


ansible/roles/runner/tasks/custom_vagrant_catalog.yml
```diff
@@ -1,18 +1,27 @@
...
+        line: "{{ custom_vagrant_catalog }} prci-automation"
+
+    - name: Set "VAGRANT_SERVER_URL" Environment variable
+      ini_file:
+        path: /etc/systemd/system/prci.service
+        section: Service
+        option: Environment
+        value: "VAGRANT_SERVER_URL='http://prci-automation'"
+        mode: 0700
+  when: custom_vagrant_catalog is defined and custom_vagrant_catalog != 'False'
```
serg-cymbaluk: The variable has `default: false`, so result of the condition `custom_vagrant_catalog != 'False'` will be always true.

netoarmando: At some point, Ansible cast that boolean to string, that's why I'm comparing with `"False"`, I believe because that's `prompt`'s behaviour, since it's expecting an input from the user.

```python
ansible-playbook test.yaml                                                                                                     

Custom vagrant catalog host (ip address) (Press enter to ignore) [False]: 

PLAY [prepare development test runners] ****************************************************************************************************

TASK [debug] *******************************************************************************************************************************
ok: [localhost] => {
    "msg": "Value False"
}

TASK [check we're not deploying against FreeIPA repo] **************************************************************************************
ok: [localhost] => {
    "custom_vagrant_catalog": false,
    "failed_when_result": false
}

PLAY RECAP *********************************************************************************************************************************
localhost                  : ok=2    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
```
```python
ansible-playbook test.yaml                                                                                                     

Custom vagrant catalog host (ip address) (Press enter to ignore) [False]: foo

PLAY [prepare development test runners] ****************************************************************************************************

TASK [debug] *******************************************************************************************************************************
ok: [localhost] => {
    "msg": "Value foo"
}

TASK [check we're not deploying against FreeIPA repo] **************************************************************************************
fatal: [localhost]: FAILED! => {
    "custom_vagrant_catalog": "foo",
    "failed_when_result": true
}

PLAY RECAP *********************************************************************************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0
```

ansible/roles/automation/box_hosting/tasks/main.yml
```diff
@@ -62,12 +67,18 @@
...
   service:
     enabled: true
     state: started
     name: gunicorn.socket
 
+- name: enable gunicorn
+  service:
+    enabled: true
+    state: started
+    name: gunicorn
```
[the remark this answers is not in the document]
netoarmando: You are right! (https://www.freedesktop.org/software/systemd/man/systemd.unit.html)
Gunicorn documentation is misleading: https://docs.gunicorn.org/en/stable/deploy.html#systemd

Issue #264, opened 2019-04-24
PR-CI is pulling data from the wrong commit


netoarmando:
(Related to #258)

Pull request https://github.com/freeipa/freeipa/pull/3018 was hanging to execute one of the tests:
![Screenshot from 2019-04-24 15-09-04](https://user-images.githubusercontent.com/217367/56686802-2fdd8c00-66ab-11e9-9a2d-ab6ec39a11a3.png)

However, that test wasn't present on the tree when [PR was opened](https://github.com/freeipa/freeipa/tree/1441d00f12dce03447c14ffe28f75e28aa7da7c3). 

Runner logs were displaying the following messages:
```bash
Apr 24 17:55:22 ad-runner-2 bash[23526]: INFO: __main__: Skipping PR#2766: can't be merged
Apr 24 17:55:23 ad-runner-2 bash[23526]: ERROR: __main__: 'NoneType' object is not subscriptable
Apr 24 17:55:24 ad-runner-2 bash[23526]: INFO: __main__: Skipping PR#3062: can't be merged
```
At that moment, PR `#3018` was the only one with pending tasks between `#2766` and `#3062`, so it was the culprit for that `ERROR` message.

That PR had only one commit (`1441d00f12dce03447c14ffe28f75e28aa7da7c3`) but PR-CI was trying to get the data from a different `sha`:
```
https://raw.githubusercontent.com/freeipa/freeipa/575ee956b4f58a160e06fb869284048b32c6020a/.freeipa-pr-ci.yaml
```
That URL was returning 200 but with not content. That's why we got that error message in the logs, raised by: https://github.com/freeipa/freeipa-pr-ci/blob/2dec5bcaa4050708333fee4163a88b2adf293adf/github/prci.py#L100-L104

The author of the PR had force-pushed some changes after that test suite was added to the master branch. Sometime during the evolution of this PR, that test suite (`fedora-29/test_installation_TestInstallMaster`) was included into `master`; test definitions couldn't be fetched from this PR, so master branch was used, based in the code below: 
https://github.com/freeipa/freeipa-pr-ci/blob/2dec5bcaa4050708333fee4163a88b2adf293adf/github/internals/entities.py#L469-L471

A complex set of events had to occur to end up into this problem, but I believe this is related to the fact that PR-CI is using the wrong commit after a force-push. 

I'm asking the author to rebase his work and try again, this should fix this pull request at least.

netoarmando:
[...]

PR #349, opened 2020-02-14
Return head commit from pull requests

netoarmando:
Github sorts commits by their 'author date', not by their order in tree, thus last commit it's not always the head of a pull request.

This returns the commit matching its `oid` (hash) with pull request's `headRefOid` value.

Last 250 (maximum allowed) commits from a given pull request must be fetched to keep the number of API requests the same as it is.

Issue: https://github.com/freeipa/freeipa-pr-ci/issues/264

Pull request used for testing this: https://github.com/netoarmando/freeipa/pull/47

---
@frasertweedale I was writing the commit message for the solution we discussed, then I decided to take another look to github's doc and I've found the `headRefOid` property. :-D

### DIFF
--- a/github/internals/gql/queries.py
+++ b/github/internals/gql/queries.py
@@ -9,6 +9,7 @@ def make_pull_requests_query(owner: Text, repo: Text) -> Dict[Text, Text]:
       nodes {
         number
         baseRefName
+        headRefOid
         mergeable
         author {
           login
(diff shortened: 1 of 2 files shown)


frasertweedale:
@netoarmando without having tested it, the approach seems correct to me.  Perhaps include in the commit message some links to the GitHub API doc i.e. https://developer.github.com/v4/object/pullrequest/#headrefoid ?  

But for sure, this is nicer than having to build the graph ourselves.

PR #359, opened 2020-04-19
Changes required to build Fedora 32+ templates

netoarmando:
- Bump rawhide release

- Replace fallocate with dd

    Due to a known bug in some filesystems where fallocate file is raising errors when trying to enable swap.

- Disable tmpfs plugin in mock    

    Build with it enabled is failing for Fedora 32 (Beta) and Rawhide (33).
    
    Stable `mock` versions as of today:
    mock-2.1-1.fc30
    mock-2.1-1.fc31
    mock-2.2-1.fc32
    mock-2.2-1.fc33

- Add tags to tasks for helping with template generation

    They can be used to when template generation need to be retried. E.g:
    ```
    $ ./create-template-box --skip-tags "upload_box,download_rawhide_box"
    ```

### DIFF
--- a/ansible/roles/builder/prepare/templates/mock-fedora-branched.cfg.j2
+++ b/ansible/roles/builder/prepare/templates/mock-fedora-branched.cfg.j2
@@ -7,6 +7,8 @@ config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
 config_opts['releasever'] = '{{ fedora_releasever }}'
 config_opts['package_manager'] = 'dnf'
 
+{# tmpfs plugin disabled for Fedora 32 and Rawhide to workaround a build failure #}
+{% if fedora_releasever < 32 %}
 # Configure tmpfs to speed up mock
 config_opts['plugin_conf']['tmpfs_enable'] = True
 config_opts['plugin_conf']['tmpfs_opts'] = {}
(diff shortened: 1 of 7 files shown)


ansible/roles/builder/prepare/templates/mock-fedora-branched.cfg.j2
```diff
@@ -7,13 +7,15 @@ config_opts['extra_chroot_dirs'] = [ '/run/lock', ]
 config_opts['releasever'] = '{{ fedora_releasever }}'
 config_opts['package_manager'] = 'dnf'
 
+{% if fedora_releasever < 32 %}
```
pvoborni, line 10: Does the tmpfs issue have some bug filed? If so we can add a comment that this is a workaround for it so that the future maintainer can check if it is still relevant and remove if not.

pvoborni:
It looks good to me - by reading. If it works for you, than we can merge it.

ansible/roles/builder/prepare/templates/mock-fedora-branched.cfg.j2
netoarmando: I didn't find anything and I don't have a clear reproducer to fill an issue myself.
This was only used to speed up the build during the generation of templates, so the impact is minimum in my opinion. Anyway I'm adding a comment to make clear that it is a workaround.

netoarmando: @pvoborni was that enough for now?

netoarmando: Marking as resolved, since it was discussed in another channel.

PR #361, opened 2020-04-29
prevent NetworkManager from updating /etc/resolv.conf

wladich:
NM periodically updates contents of /etc/resolv.conf putting
dhcp-provided DNS server on top of the list which breaks the
tests relying on DNS data provided by IPA server.

### DIFF
--- a/ansible/roles/machine/provision/tasks/main.yml
+++ b/ansible/roles/machine/provision/tasks/main.yml
@@ -75,6 +75,20 @@
     src: hosts
     dest: /etc/hosts
 
+- name: prevent NetworkManager from updating/etc/resolv.conf
+  ini_file:
+    path: /etc/NetworkManager/NetworkManager.conf
+    section: main
+    option: dns
+    value: "none"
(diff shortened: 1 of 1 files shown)


wladich:
Testing the PR on private runener

wladich:
@netoarmando 
Here is a failed run with unmodified PRCI: http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/fceefbb6-8a13-11ea-a7db-fa163e43b58c/report.html: kinit for AD user fails with message "Cannot find KDC for realm "AD.TEST" while getting initial credentials" because libkrb tries to use dns discovery for this realm and fails.

When same test runs with proposed patch, it succeeds: https://github.com/wladich/freeipa/pull/17 (the job fedora-latest/test_http_kdc_proxy is stuck, I do not know what is the reason)

netoarmando:
Deployed.

wladich:
Thank you for prompt deploy

PR #371, opened 2020-06-01
Keep only 3 boxes in the runners

netoarmando:
Current boxes are larger than they used to be and small runners can't handle much more than 4 boxes.

### DIFF
--- a/tasks/vagrant.py
+++ b/tasks/vagrant.py
@@ -211,8 +211,10 @@ class VagrantBox(object):
 
         # Do not delete Windows boxes
         linux_boxes = [x for x in all_boxes if 'windows' not in x.name]
-        if len(linux_boxes) > 4:
-            linux_boxes[0].delete_box()
+
+        # Keep only the latest 3 boxes
+        for box in linux_boxes[:-3]:
+            box.delete_box()
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


tasks/vagrant.py
```diff
@@ -211,8 +211,10 @@ def clean_last_time_used(box):
 
         # Do not delete Windows boxes
         linux_boxes = [x for x in all_boxes if 'windows' not in x.name]
-        if len(linux_boxes) > 4:
-            linux_boxes[0].delete_box()
+
+        # Keep only the latest 3 boxes
+        for box in linux_boxes[:-3]:
```
netoarmando, line 216: @f-trivino this is executed when the runner wants to download a new template, so we can keep only that template that will be downloaded and delete the ones in the runner with:
```python
for box in linux_boxes:
    box.delete_box()
```
The problem was that some runners were keeping 6 boxes (I don't know why), they are larger, we were keeping 5 and previous code was only deleting one on them.

bhavikbhavsar:
has this change been tested ?

netoarmando:
> has this change been tested?

yes, but I failed to collect its log output

bhavikbhavsar:
LGTM

PR #373, opened 2020-06-04
prevent NetworkManager from updating /etc/resolv.conf on ipa client

wladich:
NM periodically updates contents of /etc/resolv.conf putting
dhcp-provided DNS server on top of the list which breaks the
tests relying on DNS data provided by IPA server.

This patch is needed only for client machines as ipa-server-install
provides similar changes during IPA server installation.

### DIFF
--- a/ansible/roles/machine/provision/tasks/main.yml
+++ b/ansible/roles/machine/provision/tasks/main.yml
@@ -75,6 +75,22 @@
     src: hosts
     dest: /etc/hosts
 
+- when: inventory_hostname | regex_search("^client\d+$")
+  block:
+  - name: prevent NetworkManager from updating/etc/resolv.conf
+    ini_file:
+      path: /etc/NetworkManager/NetworkManager.conf
+      section: main
+      option: dns
+      value: "none"
+    register: nm_conf
+
+  - name: restart Networkmanager
+    service:
+      name: NetworkManager
+      state: restarted
+    when: nm_conf is changed
+
 - name: create /etc/resolv.conf file from template
   template:
     src: resolv.conf


wladich:
The PR was tested on private runner: https://github.com/wladich/freeipa/pull/17

Both new kdcproxy tests and existing TestBackupAndRestoreTrust are green.

netoarmando:
This breaks tests that install packages on clients, none of the current tests do that, but https://github.com/freeipa/freeipa/pull/4676 will be the first one.

One possible solution is to keep the changes you are proposing here + add a nameserver to the end of `resolv.conf`. It could be the gateway's IP address, since libvirt's network forwards DNS queries or a statically defined one, such as Google's.

fcami:
hi @netoarmando I'm not too sure.
All nameservers in resolv.conf should be "identical" in the response they give ; the resolver only jumps from one to the next on timeout, not on NXDOMAIN.
Having that gateway's IP address in a variable that could be reused at IPA install time as a forwarder in test suites would be a better solution: if a proper forwarder is configured, then package installation (and in general, resolving names from outside the IPA domain) will work even on clients.
In short, I would not fiddle with resolv.conf anywhere.

wladich:
@fcami, @netoarmando 
I have created a small test to demonstrate how name resolution works now: https://github.com/wladich/freeipa/pull/19/files: they print contents of resolv.conf and try to install some packages - before and after ipa installation.

Here are results without the patch: http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/10b3a61e-aa54-11ea-8cc2-fa163e4ad552/
And with the proposed patch: http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/0dfe8534-aa5d-11ea-8850-fa163e4ad552/

Without patch: dnf works both before and after ipa instalation. But the contents of resolv.conf are not as expected for properly configured IPA installation with DNS server: 
* it was set to point to ipa master and replica by ansible task `machine/provision : create /etc/resolv.conf`
* in TestDnf::test_dnf_client_before_install we can see it was changed by NetworkManager to point to single server - libvirt gateway
* after ipa installation resolv.conf has not changed which means we can not use DNS provided by IPA server

What happens with patch applied:
* ansible writes ipa master and replica to resolv.conf 
* dnf fails at this point as IPA DNS is not installed yet
* after ipa installation dnf works which means that DNS resolution works whil resolv.conf  still points to ipa master and replica. That is because by default tests install IPA server with DNS  configured to forward requests to libvirt gateway (more precisely to any dns from config file).

IMO the proper solution should be:
* keep the proposed patch so that NM does not change resolv.conf unexpectedly
* do not write resolv.conf during VM preparation, i.e. remove `machine/provision : create /etc/resolv.conf`
* modify tests to make them change resolv.conf on client to point to IPA servers only (only when IPA is installed with DNS) - this way the test environment will better match real world.

What @netoarmando proposes should work but with limitations: if vagrant dns is listed after IPA servers, then it will be used before IPA is intsalled (because IPA DNS will timeout). And when IPA is configured, ipa dns will be used because it is first in list.
The only problem is that according to `man resolv.conf`
>  Up to MAXNS (currently 3, see <resolv.h>) name servers may be  listed,  one  per  keyword

So of we have one master, two replicas and one vagrant provided DNS added to resolv.conf, libc will not be able to use the latter.


> Having that gateway's IP address in a variable that could be reused at IPA install time as a forwarder in test suites would be a better solution: if a proper forwarder is configured, then package installation (and in general, resolving names from outside the IPA domain) will work even on clients.

If I understand correctly, it is exactly how it is done now. The only problem is that in this case DNS resolution does not work on clients BEFOR IPA is installed.

wladich:
@abbra, @flo-renaud 
Please have a look at the proposal in previous comment.

flo-renaud:
@wladich 
I mostly agree with your points but we need to clarify some items:

- [x] _keep the proposed patch so that NM does not change resolv.conf unexpectedly_
- [ ] _do not write resolv.conf during VM preparation, i.e. remove machine/provision : create /etc/resolv.conf_
In this case will the VMs use the libvirt gateway by default or do we need to set it up?
- [x] _modify tests to make them change resolv.conf on client to point to IPA servers only (only when IPA is installed with DNS) - this way the test environment will better match real world._
This can be done in the install_xx methods defined in the IntegrationTest class, but a lot of tests override install and may need to be modified separately. It will also be tricky for tests that do install/uninstall or install master+replica: do you plan to add all the replicas to resolv.conf on the clients or only the master? Update resolv.conf when a master or replica is removed?

f-trivino:
> @wladich
> I mostly agree with your points but we need to clarify some items:
> 
>     * [x]  _keep the proposed patch so that NM does not change resolv.conf unexpectedly_
> 
>     * [ ]  _do not write resolv.conf during VM preparation, i.e. remove machine/provision : create /etc/resolv.conf_
>       In this case will the VMs use the libvirt gateway by default or do we need to set it up?
> 
>     * [x]  _modify tests to make them change resolv.conf on client to point to IPA servers only (only when IPA is installed with DNS) - this way the test environment will better match real world._
>       This can be done in the install_xx methods defined in the IntegrationTest class, but a lot of tests override install and may need to be modified separately. It will also be tricky for tests that do install/uninstall or install master+replica: do you plan to add all the replicas to resolv.conf on the clients or only the master? Update resolv.conf when a master or replica is removed?

@wladich what is the status of this MR?

wladich:
@f-trivino 
It has been temporarily postponed due to a large amount of other work. Hope to return to it soon.

wladich:
Notes: 
* NM should also be disabled on replicas, for the same reason.
* I think we should stop adding all VMs to /etc/hosts, this looks like a dirty workaround .

fcami:
- Disabling NM on replicas does not match what's in the real world.
- +1 to stopping to add all VM IP/FQDN to /etc/hosts.

abbra:
May be we should instead add NM and systemd-resolved overrides like we do on masters/replicas with DNS role?
Make sure they point to our masters and be done with it. This is a specific setup of our test environment, all we need to ensure is that the hosts are able to reach IPA DNS servers and that's all.

fcami:
@abbra : The one thing that gives me pause is for instance the locations feature, or testing clients against a specific replica. Essentially, if we pre-configure clients to point at a set of 3 IPA servers, that's not how this is done IRL. But there is no reason these specific tests could not override that particular configuration bit before running, so it's a weak argument against what you just proposed.

wladich:
@ab: that is exactly what I propose in the PR, please see the diff.

> May be we should instead add NM and systemd-resolved overrides like we do on masters/replicas with DNS role?

That is what I am proposing, please see the diff

> Make sure they point to our masters and be done with it. This is a specific setup of our test environment, all we need to ensure is that the hosts are able to reach IPA DNS servers and that's all.

This is the tricky part, I showed it in https://github.com/freeipa/freeipa-pr-ci/pull/373#issuecomment-641899414. In short: currently resolv.conf is populated with ipa masters/replicas by ansible and is immediately rewritten by NM. If we stop NM from doing it, tests will not be able to use dnf on clients before IPA server is installed. So IMO this should work following way: client machines start with initial state of resolv.conf provisioned by vagrant (i.e. pointing to libvirt gateway) and with the NM override which disables managing of resolv.conf. Then resolv.conf on client is configured to point to IPA master after packages are installed in tests and before tests are executed, i.e. somewhere around tests setup. But as @flo has noticed this requires careful changes in many test suites.

abbra:
@wladich we should not change resolv.conf. Instead, we should add an override file to NM configuration -- we already know how to do this in IPA master/replica installers, so adding this to NM (and systemd-resolved) configurations would be the right way to do it.

@fcami yes, the individual tests should have this overridden,  if they need it. However, we should provide a default environment where DNS resolution of IPA zones just works. The reverse should be an exception, rather than the rule.

wladich:
> * Disabling NM on replicas does not match what's in the real world.

I am not sure how it should be done the proper way, but what I see now:
* After installation of IPA master and replica, reslov.conf on replica has libvirt gateway as first nameserver line
* File /etc/NetworkManager/conf.d/zzz-ipa.conf does not exist on replica

So replica is using libvirt gateway as a DNS, which is wrong.

wladich:
@abbra 
if we "not change resolv.conf" on clients, it will not contain IPA master or replica. It will contain only libvirt gateway, like it does now.

abbra:
@wladich when NM owns resolv.conf, it knows how to delegate resolving functions to different locations as per its configuration. Thus, our test deployment playbooks should set up something similar to `zzz-ipa.conf` override on client and replica systems.

wladich:
We have discussed the PR with @abbra, here are the findings:
* we should not block management of resolv.conf by  NM or systemd-resolved as it does not match real usage
* systemd points resolv.conf to local DNS server called systemd-resolved which can be configured to route requests to IPA master or libvirt gateway based on the domain. Good solution
* On systems using NM we could configure it to properly manage resolv.conf same way it is done for ipa server, but that does not solve the problem of libc using maximum 3 servers.

Possible solution #1: setup dnsmasq on controller vm and point resolv.conf on clients to it (using overrides for NM and systemd-resolved). This dnsmasq should route requests based on the domain name.

Possible solution #2: Use only IPA master and libvirt gateway as DNS servers (also configure it using overrides for NM and systemd-resolved configs), i.e. do not use replicas.

flo-renaud:
I wonder if the issue isn't linked to the way we configure the forwarder in freeipa-pr-ci. When the test controller machine is configured, the ansible playbook tries to find which DNS forwarder to use based on the content of /etc/resolv.conf on the controller (see the role [provision](https://github.com/freeipa/freeipa-pr-ci/blob/master/ansible/roles/machine/provision/tasks/main.yml#L3)):
```
    - name: get DNS server from resolv.conf
      shell: awk '$1 == "nameserver" {print $2; exit}' /etc/resolv.conf
      register: dns_server

```
Then this DNS forwarder is set in the [ipa-test-config.yaml](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/8b70b850-2845-11eb-ba75-fa163e524e71/ipa-test-config.yaml) as the dns_forwarder and ipa-server-install is called with --forwarder "value of dns_forwarder=127.0.0.53". This looks wrong as we end up with our IPA DNS server using the loopback as forwarder.
I think we should modify the playbook to get the DNS resolver from the output of `resolvectl dns` when the controller is on fedora33.

PR #377, opened 2020-06-23
Allow cipherboy (Alexander Scheel) to launch PRCI tests

flo-renaud:


cipherboy:
Thanks @flo-renaud! :)

PR #380, opened 2020-06-29
Enforce AES encryption

amore17:
As Fedora 32's default crypto policy does
not support RC4 any more, thus AES encryption
enabled on AD side.

Fixes: https://pagure.io/freeipa/issue/8294


abbra:
LGTM

PR #385, opened 2020-07-22
Keep only 2 boxes

netoarmando:
Production runners require 2 boxes only (latest master and ipa-4-8 branches). Nightly runners are running close to the threshold to hit out of space issues.


netoarmando:
This was deployed to the parallel infra and it was proven effective. So, I'm merging it and deploying it to the production runners.

PR #390, opened 2020-08-27
Update job data from nightly runs

netoarmando:
Data from nightly runs were not uploaded to dynamodb because they are triggered from a fork.


f-trivino (approved):
Approved, thanks for the patch.

Issue #391, opened 2020-09-08
RFE: ability to run tests with SELinux in Enforcing mode


fcami:
Currently the nightlies are run with SELinux set to Permissive because an SELinux policy update could make all/most tests fail, breaking CI for weeks.
But the only realistic way to catch issues with SELinux policy updates is to have CI nightlies with SELinux set to Enforcing ; collecting AVCs will not work, because a dontaudit rule could break FreeIPA and leave no trace for instance.

Proposal:
- add a "selinux_enforcing: True" yaml parameter (just like "update_packages: True") 
- if selinux_enforcing is set to True, run "setenforce 1" on all VMs before the test starts

Adding more nightlies with "selinux_enforcing: True" for all tests:
- nightly_master_selinux, in sync with nightly_master
- nightly_latest_testing_selinux in sync with nightly_latest_testing
- nightly_ipa-4-8_latest_selinux (in ipa-4-8 branch), in sync with nightly_ipa-4-8_latest

fcami:
Awesome, thank you @netoarmando !

netoarmando:
Cron jobs have been enabled.

PR #392, opened 2020-09-08
Add support to change SELinux mode to enforcing

netoarmando:
Setting `selinux_enforcing: true` as an argument to a job definition will change the SELinux mode to `enforcing` during provisioning phase.

Issue #391 

Testing run: https://github.com/netoarmando/freeipa/pull/60


fcami:
LGTM, thank you @netoarmando

PR #402, opened 2020-11-17
Normalize rawhide definitions + Improve tool used to store templates

netoarmando:
- Print and compare checksum when adding new box
>     Update script used to download a box to the locally hosted Vagrant
>     catalog to compare box's sha1sum with the one availble on the original
>     registry, saving time when publishing new box templates.

- Normalize Rawhide definitions

>     * Update COPR repos for rawhide to don't use release numbers, build stage
>     was failing in recent (f34) release target.
> 
>     * Add 389 COPR repo to rawhide mock definition, this way branched and
>     rawhide definitions look the same.
> 
>     * Include disabled variables to rawhide.yml to provide an easier way to
>     check what's availble when building a new template.

### DIFF
--- a/ansible/roles/builder/prepare/templates/mock-fedora-rawhide.cfg.j2
+++ b/ansible/roles/builder/prepare/templates/mock-fedora-rawhide.cfg.j2
@@ -27,7 +27,7 @@ best=1
 
 [freeipa_copr]
 name=Copr repo @freeipa/freeipa-{{ freeipa_version }}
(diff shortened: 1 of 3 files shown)


bhavikbhavsar:
The changes looks good, however can we add information as to what was the problem we are trying to solve?

netoarmando:
> The changes looks good, however can we add information in PR description as to what was the problem we are trying to solve?

Thanks for the review, I've updated the commit messages, I hope they are good enough now.

bhavikbhavsar:
> > The changes looks good, however can we add information in PR description as to what was the problem we are trying to solve?
> 
> Thanks for the review, I've updated the commit messages, I hope they are good enough now.

Excellent! now I understand, we can merge this changes as well. LGTM

PR #403, opened 2020-11-18
Make lock mechanism quicker and clearer

netoarmando:
By removing the sleep when locking a task the time between polling the
status and setting a new one is reduced, this is important to lower the
chances of a race condition.

"Taken" is being divided into "Locked" and "Taken", that can also help us
to troubleshoot future issues.

"pending for rerun" status now includes the runner that set it and when
it was set.


netoarmando:
Test run output showing the issue and the changes proposed by this PR:
Screenshot from 2020-11-23 09-54-45

bhavikbhavsar:
LGTM

PR #405, opened 2020-12-03
Modernize spec file modifications

abbra:
FreeIPA spec file template already contains autoconf use, no need to add
another one.

Do not replace Version: line, replace @VERSION@ in IPA_VERSION
definition alone. IPA_VERSION macro is reused across the spec file.

Signed-off-by: Alexander Bokovoy <abokovoy@redhat.com>


netoarmando (approved):
LGMT, working fine against `master` and `ipa-4-8` branches.

netoarmando:
Deployed to all runners.

PR #407, opened 2020-12-07
Fix Fedora 33 DNS forwarder

netoarmando:
Fedora 33 switched its default DNS resolver to systemd-resolved, previously the task was getting the DNS server from /etc/resolv.conf, but for Fedora 33 that value is a loopback, breaking IPA installation because that address is used to define a DNS forwarder.

Testing this here: http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/f6a1b546-38be-11eb-95c6-5254000d65ad/

### DIFF
--- a/ansible/roles/machine/provision/tasks/main.yml
+++ b/ansible/roles/machine/provision/tasks/main.yml
@@ -1,8 +1,14 @@
 ---
 - block:
-    - name: get DNS server from resolv.conf
+    - name: get DNS server from resolv.conf (fedora <= 32)
       shell: awk '$1 == "nameserver" {print $2; exit}' /etc/resolv.conf
       register: dns_server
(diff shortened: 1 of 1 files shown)


flo-renaud (commented):
Hi @netoarmando 
thanks for the fix, LGTM.
Just to double-check, are we using only Fedora controllers? (otherwise the dns_server variable will never get assigned any value).

netoarmando:
> Hi @netoarmando
> thanks for the fix, LGTM.
> Just to double-check, are we using only Fedora controllers? (otherwise the dns_server variable will never get assigned any value).

Yes, controllers use the same vagrant boxes and when creating a new box users are prompted to choose which Fedora release to base that box. Pretty safe to say that PR-CI only supports Fedora and work would be necessary to support any other distro.

flo-renaud (approved):
@netoarmando thanks for the confirmation, you can merge

Issue #410, opened 2020-12-12
DNS forwarder value needs to be stripped of trailing spaces in rawhide


flo-renaud:
When running the PRCI tests on rawhide, the dns_forwarder is extracted using `resolvectl status` command (https://github.com/freeipa/freeipa-pr-ci/blob/master/ansible/roles/machine/provision/tasks/main.yml#L18) but this results in a variable containing trailing spaces.

See for instance the PR #[586](https://github.com/freeipa-pr-ci2/freeipa/pull/586) where the [test_webui_users](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/25a104c0-3c91-11eb-a77b-fa163ed58e5d/) test fails. The generated [ipa-test-config.yaml](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/25a104c0-3c91-11eb-a77b-fa163ed58e5d/ipa-test-config.yaml) contains trailing spaces for the dns_forwarder value and this results in ipa-server-install failure visible in [runner.log.gz](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/25a104c0-3c91-11eb-a77b-fa163ed58e5d/runner.log.gz):
```
2020-12-12 15:52:23,733    DEBUG  TASK [machine/provision_ipaserver : install freeipa] ***************************
2020-12-12 15:52:26,840    DEBUG  fatal: [master]: FAILED! => {
2020-12-12 15:52:26,840    DEBUG      "changed": true,
2020-12-12 15:52:26,841    DEBUG      "cmd": "ipa-server-install -U -n ipa.test -r IPA.TEST -p Secret.123 -a Secret.123 --setup-dns --setup-kra --forwarder='192.168.122.1            '\n",
2020-12-12 15:52:26,841    DEBUG      "delta": "0:00:02.142841",
2020-12-12 15:52:26,841    DEBUG      "end": "2020-12-12 15:52:26.301897",
2020-12-12 15:52:26,841    DEBUG      "rc": 2,
2020-12-12 15:52:26,842    DEBUG      "start": "2020-12-12 15:52:24.159056"
2020-12-12 15:52:26,842    DEBUG  }
2020-12-12 15:52:26,842    DEBUG  
2020-12-12 15:52:26,842    DEBUG  STDERR:
2020-12-12 15:52:26,842    DEBUG  
[...]

PR #411, opened 2020-12-14
Fix resolvectl task returning white spaces

netoarmando:
Replace task command to fetch DNS sever from `resolvectl dns` command, this will not return whitespaces.
Execute `grep -m1` to limit the number of lines to 1.
Execute `tr` to remove possible spaces.

Issue https://github.com/freeipa/freeipa-pr-ci/issues/410

---

Test runs:
- `fedora-32/test_webui_users`: http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/7f901ebe-3e24-11eb-9129-5254000d65ad
- `fedora-33/test_webui_users`: http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/ccc97ac6-3e2a-11eb-9129-5254000d65ad
- `fedora-rawhide/test_webui_users`" http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/d0da5f62-3e30-11eb-9129-5254000d65ad

### DIFF
--- a/ansible/roles/machine/provision/tasks/main.yml
+++ b/ansible/roles/machine/provision/tasks/main.yml
@@ -6,7 +6,7 @@
       when: ansible_distribution == 'Fedora' and ansible_distribution_version is version('32', '<=')
 
     - name: get DNS server from resolvectl (fedora >= f33)
-      shell: "resolvectl status | grep 'Current DNS Server' | awk -F': ' '{print $2}'"
+      shell: "resolvectl dns | grep -m1 ': [1-9]' | awk -F': ' '{print $2}' | cut -d' ' -f 1"
       register: dns_server_resolvectl
       when: ansible_distribution == 'Fedora' and ansible_distribution_version is version('33', '>=')
 
@@ -102,7 +102,7 @@
 - name: set hostname
   shell: "hostnamectl set-hostname {{ inventory_hostname }}.ipa.test"
 
-# Change selinux state if `selinux_enforcing: true` is set in test suite definition 
+# Change selinux state if `selinux_enforcing: true` is set in test suite definition
 - name: set selinux to enforcing
   selinux:
     policy: targeted


ansible/roles/machine/provision/tasks/main.yml
```diff
@@ -6,7 +6,7 @@
       when: ansible_distribution == 'Fedora' and ansible_distribution_version is version('32', '<=')
 
     - name: get DNS server from resolvectl (fedora >= f33)
-      shell: "resolvectl status | grep 'Current DNS Server' | awk -F': ' '{print $2}'"
+      shell: "resolvectl dns | grep -m1 'Link' | awk -F': ' '{print $2}' | tr -d '[:space:]'"
```
flo-renaud, line 9: I tried this cmd on a 1minutetip machine but this would return a mangled output:
```
# resolvectl dns
Global:
Link 2 (eth0): 10.11.5.19 10.5.30.45
# resolvectl dns | grep -m1 'Link' | awk -F': ' '{print $2}' | tr -d '[:space:]'
10.11.5.1910.5.30.45
```
There are multiple issues:
- we can't assume that the DNS server is always configured at the link level, it could also be set for "Global: "
- multiple servers can be listed in each line

Something like `resolvectl dns | grep -m1 ': [1-9]' | awk -F': ' '{print $2}' | cut -d' ' -f 1` should work (discard lines without any IP address and consider only the first one, extract the content after ': ', split with space and take only the first one.

netoarmando: Thanks @flo-renaud! I wasn't expecting the link to return more than one address.

What I don't understand about the this issue is why the [ipa-test-config.yaml](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/25a104c0-3c91-11eb-a77b-fa163ed58e5d/ipa-test-config.yaml) you linked shows 2 addresses (one from Global another from Link supposedly) because the one from Global should have been set by a task called later [`create /etc/resolv.conf file from template`](https://github.com/freeipa/freeipa-pr-ci/blob/77bb1540f8157558b65ec6a718162e8d1d3982d4/ansible/roles/machine/provision/tasks/main.yml#L92-L95). 

I failed to reproduce that.

The behaviour below shouldn't affect `ipa-test-config.yaml` generation, but If that happens like the above, then the global IP might be 'master's address, not runner's IP. 

These are the outputs from `resolve dns` in various stages:
* Before executing any playbook:
  ```
  $ resolvectl dns
  Global:
  Link 2 (eth0): 192.168.124.1
  ```
* After updating /etc/resolv.conf:
  ```
  resolvectl dns
  Global: 192.168.124.252
  Link 2 (eth0): 192.168.124.1
  ```
* Few minutes after executing that (because NM updated the `resolv.conf`)
  ```
  resolvectl dns
  Global: 127.0.0.1
  Link 2 (eth0): 192.168.124.1
  ```

f-trivino (approved):
LGTM. The command suggested by @flo-renaud  works for me. I saw you updated the PR message as well, I'm approving it.

PR #416, opened 2021-01-21
Replace ipa-4-8 with ipa-4-9 nightly jobs

netoarmando:
Disable nightly runs for ipa-4-8 branch and enable ipa-4-9 to replace them.


netoarmando:
@flo-renaud Should we remove ipa-4-6 nightlies as well?

flo-renaud:
> @flo-renaud Should we remove ipa-4-6 nightlies as well?

@netoarmando no, please keep them. The branch is still in use for RHEL 7.9 z stream but with a small difference (RHEL 7.9 is python2 only, while fedora 27 supports python2 and python3), and I'd rather keep on testing both variants.

flo-renaud (approved):
@netoarmando Thanks for the PR, LGTM

Issue #422, opened 2021-03-02
Tests timeout with "No free disk space" error message


wladich:
Timeout can occur

* during provisioning: [test_ipahealthcheck_adtrust](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/f36f1706-7ac7-11eb-a446-fa163e5ccdc8/), [test_sssd](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/4bff1526-7b02-11eb-a349-fa163e0817f7/), [another test_sssd](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/d0804ef0-6fb8-11eb-9f76-fa163eeafbeb/), [test_trust](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/f718c300-770e-11eb-8f9f-fa163e04929f/)
```
2021-03-01 20:36:05,264    DEBUG  TASK [utils : write swap file] *************************************************
2021-03-01 21:14:36,632 CRITICAL  Terminating execution, runtime exceeded 4800s
2021-03-01 21:14:36,646 CRITICAL  No free disk space
2021-03-01 21:14:36,679    DEBUG  Process "vagrant provision" exited with error code -9
```

* during test execution: [test_trust](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/8dcaf918-78a3-11eb-b0d3-fa163e04929f/)
```
2021-02-27 03:47:01,510    DEBUG  [ipatests.pytest_ipa.integration.host.Host.client0.cmd229] tar: /var/log/ipaserver-kra-install.log: Warning: Cannot stat: No such file or directory
2021-02-27 03:47:01,524    DEBUG  [ipatests.pytest_ipa.integration.host.Host.client0.cmd229] Exit code: 0
2021-02-27 03:47:01,527    DEBUG  [ipatests.pytest_ipa.integration.host.Host.client0.IPAOpenSSHTransport] GET /tmp/tmp.jnquSKZ6rD
2021-02-27 03:47:01,528    DEBUG  [ipatests.pytest_ipa.integration.host.Host.client0.cmd230] RUN ['cat', '/tmp/tmp.jnquSKZ6rD']
2021-02-27 04:59:01,916 CRITICAL  Terminating execution, runtime exceeded 9000s
[...]

PR #424, opened 2021-03-04
Fix autocleaner

wladich:
Fixing two issues in autocleaner.py:
* Crash while removing old jobs dirs (#422)
* `--jobs_dir_exp` option was not respected, all jobs dirs were removed


netoarmando (approved):

netoarmando:
Deployed to all runners.

PR #432, opened 2021-06-07
Make all Cron entries unique

netoarmando:
After 442fe16 only the last entry of each with same name was considered, so a unique name must be set.

Entry `testing_master_previous` was schuduled to Thursdays.


netoarmando:
@flo-renaud I've deployed using the changes from this, but feel free to suggest better suffixes ('monday', 'tuesday' are a bit ugly).

flo-renaud:
@netoarmando 
if the PR has a different we loose the capacity to compare the test results with https://ci.freeipa.org/nightlies.
Would it be possible to have a definition like:
```
nightly_jobs:
  - job: job1 
    name: testing_master_latest
    weekdays: "1"
    hour: "23"
    minute: "00"
    flow: "ci"
    branch: "master"
    prci_config: "nightly_latest.yaml"
    reviewer: wladich
  - job: job2
    name: testing_master_latest
    weekdays: "3,5"
    hour: "23"
    minute: "00"
    flow: "ci"
    branch: "master"
    prci_config: "nightly_latest.yaml"
    reviewer: flo-renaud

```
This way the primary key for each item is "job" (but the value is not used anywhere), and we can keep the name?

netoarmando:
@flo-renaud I've updated the task name to handle that better, now PRs can have the same title. However, that means Monday and Thurday's PR will be closed on the next day.

flo-renaud:
> @flo-renaud I've updated the task name to handle that better, now PRs can have the same title. However, that means Monday and Thurday's PR will be closed on the next day.

I believe that was already the case, when a new PR with the same name is created, the previous one is closed. LGTM.

flo-renaud (approved):

PR #438, opened 2021-08-05
Clean up some old names from the allowed list

rcritten:


rcritten:
FWIW I was going to add ckelleyRH to the allowed list and noticed a bunch of old contributors names. I'll recommend him in a second PR once this merges.

netoarmando (approved):

netoarmando:
Thanks @rcritten!

PR #440, opened 2021-08-17
Adding new ADusers and groups required for auto-private-group tests

menonsudhir:
1. Create posix user create a user without gid
2. Create posix user with gid but no corresponding group.
3. Create nonposix user with a primaryGroupID that is not the default one

Signed-off-by: Sudhir Menon <sumenon@redhat.com>

### DIFF
--- a/ansible/roles/windows/ipa-ad-data/tasks/ad-root.yml
+++ b/ansible/roles/windows/ipa-ad-data/tasks/ad-root.yml
@@ -13,6 +13,11 @@
     attributes:
       gidNumber: 10048
 
+- name: testgroup1
+  win_domain_group:
+    name: testgroup1
+    scope: global
+
 - name: A test user with posix attributes defined
   include_role:
     name: windows/ipa-ad-data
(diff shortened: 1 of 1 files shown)


amore17:
Please do rebase correctly.

amore17:
Testing in : https://github.com/amore17/freeipa/pull/6

amore17 (commented):
LGTM

amore17:
@flo-renaud Test is green. If you need any changes please review this.

ansible/roles/windows/ipa-ad-data/tasks/ad-root.yml
```diff
@@ -13,6 +13,13 @@
     attributes:
       gidNumber: 10048
 
+- name: testgroup1
+  win_domain_group:
+    name: testgroup1
+    scope: global
+    attributes:
+      gidNumber: 55555555
```
flo-renaud, line 21: For consistency I would use a gidNumber that is roughly the same range as the existing users/groups (100xx are used for ad-root, 101xx for ad-child, 102xx for ad-tree).
This will also reduce the risk of having ids outside of the ranges.

But the real question is: do we need a gidNumber for this group? It is used in the non-posix scenario with nonposixuser1, and IMO the gidNumber is not needed.

flo-renaud (commented):
Hi @menonsudhir 
Can you also update the wiki page https://www.freeipa.org/page/V4/AD_configuration_for_testing ?

ansible/roles/windows/ipa-ad-data/tasks/ad-root.yml
menonsudhir: @flo-renaud, thank you for the review, i did remove the gidNumber from the above group as you mentioned.
However i am unable to authenticate to freeipa.org with fedora account for now. Will check this and update the page accordingly.

flo-renaud (approved):
Hi @menonsudhir 
thanks for the update, the changes LGTM and can be merged.

netoarmando:
Changes have been deployed to all runners.

PR #452, opened 2022-02-22
Add support for FIPS mode

miskopo:
Introduce fips bool variable and its handling, implement FIPS setup to          
provisioning playbooks. Include check to ensure that FIPS mode is                
initialized correctly prior to launching the test phase.                           
                                                                                   
Add support for reloading all machines from vagrant's side (this halts             
all the machines, starts them up and reestablishes shared folder).                 
Reload can be triggered by creating a REBOOT_READY file in the vagrant            
shared folder before the test phase (either in provisioning or upping             
phase).                                                                           
                                                                                
Signed-off-by: Michal Polovka <mpolovka@redhat.com>

PR #453, opened 2022-03-09
Fix deprecated import shortcut

netoarmando:
Python 3.8.7 was already printing warning about that when import Callable from collections:

```
DeprecationWarning: Using or importing the ABCs from 'collections'
instead of from 'collections.abc' is deprecated since Python 3.3,
and in 3.9 it will stop working.
```

Fedora 35 runners have Python 3.10 so this is needed to support PR-CI.

### DIFF
--- a/tasks/common.py
+++ b/tasks/common.py
@@ -1,11 +1,11 @@
 import abc
-import collections
 import errno
 import logging
 import os
 import signal
 import subprocess
 import threading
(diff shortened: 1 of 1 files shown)


netoarmando:
@miskopo I'm deploying this change + https://github.com/freeipa/freeipa-pr-ci/pull/452 on your private runner right now.

miskopo:
This change looks good to me, ACK

PR #462, opened 2022-05-18
Wait 30 seconds after vm is rebooted

netoarmando:
Otherwise, vagrant package might generate broken vm, without ssh.


f-trivino (approved):
LGTM

PR #463, opened 2022-05-23
Set custom COPR repo via test definition files

netoarmando:
Replace defined-during-image-build definitions of COPR repos with a "copr" and "enable_testing_repo" vars that can be set in the test definitions file.

E.g:

```
fedora-latest/build:
  requires: []
  priority: 100
  job:
    class: Build
    args:
      enable_testing_repo: true
      copr: '@pki/master'
      git_repo: '{git_repo}'
      git_refspec: '{git_refspec}'
      template: &ci-master-latest
        name: freeipa/ci-master-f35
        version: 0.0.5
      timeout: 1800
      topology: *build
```

Also, the "build" job needs to set those variables too.

And `update_packages` must be set to `True` when setting `enable_testing_repo` or `copr`; otherwise, packages available on those repos wouldn't be updated.

### DIFF
--- a/ansible/roles/builder/build/tasks/create_rpms.yml
+++ b/ansible/roles/builder/build/tasks/create_rpms.yml
@@ -2,6 +2,8 @@
 - name: create srpm
   shell: |
     mock --buildsrpm \
+      {% if copr is defined and copr %}--addrepo 'https://download.copr.fedorainfracloud.org/results/{{ copr }}/fedora-$releasever-$basearch/'{% endif %}
+      {% if enable_testing_repo is defined and enable_testing_repo|bool %}--enablerepo updates-testing{% endif %}
       --spec /root/rpmbuild/SPECS/freeipa.spec \
       --sources /root/rpmbuild/SOURCES/freeipa-{{ build_version }}.tar.gz \
       --result /root/rpmbuild/SRPMS/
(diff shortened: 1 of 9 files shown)


netoarmando:
Testing run: https://github.com/netoarmando/freeipa/pull/70

f-trivino (commented):
I think setting `copr` or `enable_testing_repo` without `update_packages` doesn't make sense. The later is easy-to-forget, and must be added to every single job definition by hand, right?. If that's the case, what about this simplification?

- `update_packages` is automatically enabled when `copr` and/or `enable_testing_repo` is set

and if the above scenario makes sense then maybe we can honor update_packages when it is set to False by intention.


I think it would be good if we document the new feature and we add an example (similar to [1]) where we indicate how to:

1) set the new vars within the build job definition
2) how to set update_packages

as part of(?): https://github.com/freeipa/freeipa-pr-ci/blame/master/doc/README.md#L274


What about adding a commit to this PR where 389ds, pki, testing templates are updated?

----
I've verified that your test is working fine and all scenarios are installing the right packages.

[1] https://github.com/netoarmando/freeipa/pull/70/commits/1c61fbf046464df4b3697ff2f2017d509db3adfc

netoarmando:
> and if the above scenario makes sense then maybe we can honor update_packages when it is set to False by intention.

I agree with you, but I decided not to follow that because currently the custom images "require" the `update_packages` set to True in their definitions, so I left the same behavior.

> I think it would be good if we document the new feature and we add an example (similar to [1]) where we indicate how to..

I forgot that, updating README to include the details.

f-trivino:
> I agree with you, but I decided not to follow that because currently the custom images "require" the `update_packages` set to True in their definitions, so I left the same behavior.

right, I missed that, so let's keep current implementation. Let the documentation to do it's job then :)

f-trivino:
@netoarmando I've reviewed the README content. LGTM!.

netoarmando:
Deployed to all runners.

Issue #467, opened 2022-06-26
Rawhide image is missing pytest-html plugin


flo-renaud:
The rawhide image is missing pytest-html plugin. As a consequence, the PRCI nightly runs are failing on rawhide, see for instance PR #[1815](https://github.com/freeipa-pr-ci2/freeipa/pull/1815) with the test `fedora-rawhide/automember` ([Details](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/a889504e-f47e-11ec-826e-fa163e21c439/)).

The [runner log](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/a889504e-f47e-11ec-826e-fa163e21c439/runner.log.gz) shows the following error:
```
2022-06-25 12:15:00,560     INFO  Executing: Process "vagrant ssh -c IPATEST_YAML_CONFIG=/vagrant/ipa-test-config.yaml ipa-run-tests test_integration/test_automember.py --verbose --logging-level=debug --logfile-dir=/vagrant/ --html=/vagrant/report.html --junit-xml=/vagrant/junit.xml"
2022-06-25 12:15:05,839    DEBUG  ERROR: usage: -c [options] [file_or_dir] [file_or_dir] [...]
2022-06-25 12:15:05,840    DEBUG  -c: error: unrecognized arguments: --html=/vagrant/report.html
2022-06-25 12:15:05,840    DEBUG    inifile: None
2022-06-25 12:15:05,840    DEBUG    rootdir: /usr/lib/python3.11/site-packages/ipatests
```

The option `--html` is provided by the plugin pytest-html but it looks like the plugin is missing on the rawhide image.

It's a puzzling issue as the image is not new (same image was used the previous week and did not trigger the problem). Maybe something changed in the pytest dependencies and the plugin is not installed by default?

abbra:
Rawhide is using Python 3.11 now. It looks like pytest-html was never packaged in Fedora so it came through some other source (PIP? COPR?)

flo-renaud:
The new rawhide box has been uploaded and PRCI is now using the latest box version, as can be seen in PRCI #[1832](https://github.com/freeipa-pr-ci2/freeipa/pull/1832). Marking as fixed.

PR #468, opened 2022-06-28
Fix COPR and Python dependencies in Fedora Rawhide

netoarmando:
#### Fix COPR repo address in mock when in rawhide boxes

> When running rawhide Ansible's variable `ansible_distribution_release` is empty, it was expected to be `Rawhide`. So, instead of using that this change is adding a task to check the content of `/etc/fedora-release` file.

#### Install latest pytest-html

> Old 1.22.x version was Python 2.7 compatible.

Issue: https://github.com/freeipa/freeipa-pr-ci/issues/467

---

This has been tested here: https://github.com/netoarmando/freeipa/pull/71

### DIFF
--- a/ansible/roles/builder/build/tasks/create_rpms.yml
+++ b/ansible/roles/builder/build/tasks/create_rpms.yml
@@ -1,8 +1,15 @@
 ---
+- name: check if target system is running a branched release
+  command:
+    cmd: 'grep -qvi rawhide /etc/fedora-release'  # fails when rawhide
+  register: target_is_branched
+  changed_when: false
+  ignore_errors: true
+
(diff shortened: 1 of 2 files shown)


f-trivino (approved):
LGTM. @netoarmando thanks for the patch, I have verified the rawhide copr URL is something like:

https://download.copr.fedorainfracloud.org/results/ftrivino/krb5-ftrivino/fedora-rawhide-x86_64/

and that the "grep -qvi rawhide /etc/fedora-release" command fails when running in rawhide.

flo-renaud (approved):
@netoarmando 
Thanks for the PR, LGTM

PR #475, opened 2022-07-22
Add Erik Belko (b3lix) to the whitelist

miskopo:
Add a new contributor to the whitelist.

Signed-off-by: Michal Polovka <mpolovka@redhat.com>


miskopo:
@b3lix do you agree with this change?

b3lix:
ACK, i agree.

netoarmando (approved):

PR #480, opened 2022-09-14
Added user with possix attribute.

amore17:
Added user with possix attributes defined.
    
For RFE Aduser is required with possix attribute.
Where users uid and gid is same and group has info attribute defined.
Related : https://bugzilla.redhat.com/show_bug.cgi?id=2062689#c6

Signed-off-by: Anuja More <amore@redhat.com>

### DIFF
--- a/ansible/roles/windows/ipa-ad-data/tasks/ad-root.yml
+++ b/ansible/roles/windows/ipa-ad-data/tasks/ad-root.yml
@@ -6,6 +6,14 @@
     attributes:
       gidNumber: 10047
 
+- name: mytestgroup with info attribute
+  win_domain_group:
+    name: mytestgroup
+    scope: global
+    attributes:
+      gidNumber: 10055
+      info: mytestuser
+
 - name: group@group
   win_domain_group:
     name: group@group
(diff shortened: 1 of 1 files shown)


f-trivino (changes requested):
I would also add more details to the commit message.

amore17:
Tested in : https://github.com/amore17/freeipa/pull/14

flo-renaud:
@amore17 nitpick but the commit msg mentions possix (with 2 s) while it should be posix (single s).
Otherwise LGTM.
Can you update the page https://www.freeipa.org/page/V4/AD_configuration_for_testing with the new user/group? Thanks.

amore17:
> @amore17 nitpick but the commit msg mentions possix (with 2 s) while it should be posix (single s). Otherwise LGTM. Can you update the page https://www.freeipa.org/page/V4/AD_configuration_for_testing with the new user/group? Thanks.

Page is updated https://www.freeipa.org/page/V4/AD_configuration_for_testing#On_forest_root_AD

flo-renaud (approved):
Thanks, ack

netoarmando:
Deployed to all runners.

PR #483, opened 2022-12-01
Increase controller memory

netoarmando:
The `dnf install` commands were consuming more memory and failing the related tasks.
More info: https://github.com/freeipa-pr-ci2/freeipa/pull/2228#issuecomment-1328874034

Controller now have the same amount as client vms.

The `memory` property was also updated/added to all templates. 
Later, the definitions at the freeipa repo must be updated.
However, based on the numbers here, nothing should change. Those numbers, memory and cpu, are used to check if a job can be run. Since we have only two types of runners, jobs will fall in the same sized runner as now.

Small runners have, in average, 6.9Gi free when in idle; and big runners 14Gi.

I'm going to deploy this right now to the nightly runners and re-run some f36 PR.

### DIFF
--- a/templates/vagrantfiles/Vagrantfile.ad_master
+++ b/templates/vagrantfiles/Vagrantfile.ad_master
@@ -6,7 +6,7 @@
 # ad_master: &ad_master
 #   name: ad_master
 #   cpu: 4
-#   memory: 12000
+#   memory: 8096
 
 Vagrant.configure(2) do |config|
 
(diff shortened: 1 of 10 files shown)


netoarmando:
Rerun of nightly tests after this: https://github.com/freeipa-pr-ci2/freeipa/pull/2239

flo-renaud (approved):
@netoarmando 
thanks for the fix, LGTM. Big thanks for also updating the recommended topology in the comments.

PR #487, opened 2023-04-26
fix: open_close_pr to re-try PR creation on error

pvoborni:
Rarely the open_close_pr receives error from GitHub when opening a PR with message, e.g.:

"""
A pull request already exists for freeipa-pr-ci2:testing_master_latest """

The assumtion is that it might be caused by a race condition between deletion of branch of previous PR and GitHub not closing the previous PR fast enough.

This fix adds a delay before opening the MR to try to lower the probability of this race condition + it tries the opening up to 3 times.

The fix also moves out assignment of reviewer and deletion of local branch out of opening of PR to more clearly separate concerns of the individual methods.

### DIFF
--- a/github/open_close_pr.py
+++ b/github/open_close_pr.py
@@ -12,6 +12,7 @@ from git import Repo
 from urllib.parse import urljoin
 import subprocess
 import requests
+import time
 
 
 REF_FORMAT = 'refs/heads/'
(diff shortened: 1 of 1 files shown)


Tiboris (approved):
Thanks for the update @pvoborni.

The code looks clean. What eyes do not like a bit is the repetition of the workflow code in two methods: `open_template_pr` and `open_nightly_pr`. Anyways it just seems right.

Thanks, Approving.

PR #488, opened 2023-05-06
fix: remove unsupported Ansible parameter 'warn'

pvoborni:
'warn' parameter for 'args' was deprecated and removed in newer Ansible verions (e.g. the one on Fedora 37+).

Removal is required for successful run of runner deployment playbooks.


flo-renaud (approved):
@pvoborni 
thanks for the PR, LGTM.

PR #493, opened 2023-11-09
Add Fedora 39 vars and bump rawhide + some fixes to make it build

pvoborni:
**Add Fedora 39 vars and bump rawhide**

As a prep for creating new boxes for ipa-4-11 and master branches.

adjustments for building f39 ipa box

**Applying the workaround did not work, thus it was commented out.**

+ some cleanup

**box uploader to run on newer Python**

collections.Mapping moved to collections.abc module thus the code needed
to be adjusted otherwise it crashed.

### DIFF
--- a/ansible/create_template_box.yml
+++ b/ansible/create_template_box.yml
@@ -6,7 +6,7 @@
     - name: fedora_version
       prompt: fedora
       private: false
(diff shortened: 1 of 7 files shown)


abbra:
LGTM

Issue #495, opened 2024-01-15
autocleaner fails to clean runners


flo-renaud:
When deploying runners for freeipa-pr-ci, the installation also sets up a cron job to periodically clean up unused vagrant boxes and delete directories for the runs from /root/freeipa-pr-ci/jobs/

With the update to fedora 39 runners, the runners are using ruamel.yaml 0.18.5 which deprecates the call to ruamel.yaml.safe_load (see https://yaml.readthedocs.io/en/latest/api/).
The result is that the autocleaner.py script fails and the filesystem grows until there is no space left:
```
# python3 /root/freeipa-pr-ci/autocleaner.py --jobs_dir_exp 3
PRCI not active, stopping systemd service...
Clearing journalctl, retain the past two days...
Vacuuming done, freed 0B of archived journals from /run/log/journal.
Vacuuming done, freed 0B of archived journals from /var/log/journal.
Vacuuming done, freed 0B of archived journals from /var/log/journal/fad0fafeb0f64e4490c092848a6f90e5.
PRCI service stopped...
Started auto-cleaning process...
Checking if box freeipa/ci-ipa-4-10-f37    (libvirt, 0.0.2) is used
Process Process-1:
Traceback (most recent call last):
  File "/usr/lib64/python3.12/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/usr/lib64/python3.12/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/root/freeipa-pr-ci/autocleaner.py", line 372, in run
    if not box.is_box_used:
           ^^^^^^^^^^^^^^^
  File "/root/freeipa-pr-ci/autocleaner.py", line 310, in is_box_used
    def_files = prci_def.get_prci_def_files()
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/freeipa-pr-ci/autocleaner.py", line 232, in get_prci_def_files
    get_gh_token())})
    ^^^^^^^^^^^^^^
  File "/root/freeipa-pr-ci/autocleaner.py", line 185, in get_gh_token
    return load_yaml(PRCI_CONFIG)['credentials']['token']
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/root/freeipa-pr-ci/autocleaner.py", line 91, in load_yaml
    return ruamel.yaml.safe_load(yml_file)
[...]

PR #496, opened 2024-01-15
autocleaner: replace calls to deprecated ruamel.yaml.safe_load method

flo-renaud:
The method ruamel.yaml.safe_load(yml_file) has been deprecated with ruamel.yaml 0.15.0. Use instead:
  yaml = YAML(typ='safe', pure=True)
  yaml.load(...)

Fixes: https://github.com/freeipa/freeipa-pr-ci/issues/495

### DIFF
--- a/autocleaner.py
+++ b/autocleaner.py
@@ -88,7 +88,8 @@ def load_yaml(yml_path):
     """
     try:
         with open(yml_path) as yml_file:
-            return ruamel.yaml.safe_load(yml_file)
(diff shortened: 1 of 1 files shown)


flo-renaud:
@pvoborni could you review and deploy this fix on the existing runners? Thanks

pvoborni (approved):
LGTM, and looks to work.

pvoborni:
The fix was deployed on both infrastructures.

Issue #498, opened 2024-01-29
Nightly tests for PKI are not using the latest pki version


flo-renaud:
The nightly test pipeline for PKI is using the copr repo @pki/master. As of today, the version of PKI available in this repository is 11.5.0 while fedora 39 is only shipping 11.4.3-2.

The issue is that the pipeline is executed with 11.4.3-2 and not 11.5.0 (see for example this week's pipeline in PR#[3349](https://github.com/freeipa-pr-ci2/freeipa/pull/3349) and the test `dns_locations` with the following installed package [list](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/86eea19c-be2a-11ee-b015-fa163e04e861/installed_packages/installed_packages_master.log.gz): `dogtag-pki-server-11.4.3-2.fc39.1.noarch`

The nightly tests are executed using vagrant images like this [one](https://github.com/freeipa/freeipa-pr-ci/blob/master/templates/vagrantfiles/Vagrantfile.master_1repl). The vagrant image is built with a fixed PKI version, but the provisioning [instructions](https://github.com/freeipa/freeipa-pr-ci/blob/a49aeecc4750cd395d9ac9ccc2d8d0a338322137/templates/vagrantfiles/Vagrantfile.master_1repl#L52) are using an ansible playbook that should update all the packages:
[ansible-provision.yml](https://github.com/freeipa/freeipa-pr-ci/blob/master/ansible/provision.yml) is installing the role `machine/provision`, which calls dnf update * in this [block](https://github.com/freeipa/freeipa-pr-ci/blob/a49aeecc4750cd395d9ac9ccc2d8d0a338322137/ansible/roles/machine/provision/tasks/main.yml#L51-L55)

The issue happens because the upgrade of pki requires `--allowerasing`:
```
# rpm -qa dogtag-pki-ca
dogtag-pki-ca-11.4.3-2.fc39.1.noarch
# dnf copr enable -y @pki/master
# dnf upgrade dogtag-pki-ca
Copr repo for master owned by @pki                                                                                                 30 kB/s | 1.5 kB     00:00    
Dependencies resolved.

 Problem: problem with installed package dogtag-tomcatjss-8.4.1-2.fc39.noarch
[...]

PR #499, opened 2024-01-29
machine/provision role: force pki upgrade using --allowerasing

flo-renaud:
The nightly tests are not using the latest pki version because the upgrade from pki 11.4.3 to 11.5.0 requires dnf update --best --allowerasing.

Force the upgrade of pki in the machine/provision role using those options when updates_packages is set to true.

Fixes: https://github.com/freeipa/freeipa-pr-ci/issues/498

### DIFF
--- a/ansible/roles/machine/provision/tasks/main.yml
+++ b/ansible/roles/machine/provision/tasks/main.yml
@@ -54,6 +54,14 @@
     state: latest
   when: update_packages is defined and update_packages
 
+# Workaround for pki upgrade issue
+- name: update pki packages
+  dnf:
+    name: 'dogtag-pki-server'
+    state: latest
+    allowerasing: true
+  when: update_packages is defined and update_packages
+
 - name: update pip packages
   pip:
     executable: pip3


flo-renaud:
Some notes related to this patch:
- by default the dnf ansible module sets `nobest: false` (see [doc](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/dnf_module.html#parameter-nobest)) which is equivalent to always adding `--best`
- I initially wanted to modify the playbook and replace the block:
```
- name: update packages
  dnf:
    name: '*'
    state: latest
  when: update_packages is defined and update_packages
```
with
```
- name: update packages
  dnf:
    name: '*'
    state: latest
    allowerasing: true
  when: update_packages is defined and update_packages
```
but there is a bug in the ansible dnf module (https://github.com/ansible/ansible/issues/82616), and the allowerasing option does not work if `name: "*"` is specified to target all the packages. That's why I added a separate block targetting only the PKI packages.

The changes have been deployed on my personal runner and can be seen in this PR: https://github.com/flo-renaud/freeipa/pull/46
The installed packages [list](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/4ab22844-bc53-11ee-9fbc-fa163e8ed6f9/installed_packages/installed_packages_master.log.gz) shows the expected pki version: `dogtag-pki-server-11.5.0-0.2.alpha5.20240125160718UTC.2268461f.fc39.noarch`

pvoborni (approved):
LGTM

PR #507, opened 2024-09-06
Provision machines: fix hostname logic

flo-renaud:
When the test machines are provisioned, the hostname must be set (otherwise the default hostname "localhost" is used.

The current logic relies on the inventory hostname being either master / client / replica / controller / trustedmaster / trustedclient / trustedreplica but depending on the Vagrant file the inventory hostname can also be client0 replica2 etc...

In this case the hostname is not set.

Change the logic to set the hostname to xx.ipa.test if the inventory_hostname does not contain trustedxx and to xx.trustedipa.test if the inventory_hostname contains trustedxx.

### DIFF
--- a/ansible/roles/machine/provision/tasks/main.yml
+++ b/ansible/roles/machine/provision/tasks/main.yml
@@ -117,11 +117,11 @@
 # workaround for https://github.com/ansible/ansible/issues/19814
 - name: set hostname
   shell: "hostnamectl set-hostname {{ inventory_hostname }}.ipa.test"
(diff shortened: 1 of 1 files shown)


flo-renaud:
Hi @amore17 
could you review this patch? It should fix the issue we see with test_authconfig for instance in https://github.com/freeipa/freeipa/pull/7528

f-trivino (approved):
@flo-renaud thanks for fixing this regression. I've tested the regex using a dummy playbook locally, worksforme, approving, merging, redeploying.

PR #508, opened 2024-09-06
ipa-test-config generation: fix when trusted_domain is not defined

flo-renaud:
The template for generating ipa-test-config.yaml should check if the variable trusted_domain is defined.
Currently there is a condition
{% if trusted_domain %}
that should be replaced with
{% if trusted_domain is defined and trusted_domain %}

otherwise the template incorrectly adds a trusted ipa domain with no host:
- hosts: name: trustedipa.test type: TRUSTED_IPA


flo-renaud:
Testd in my personal runner with PR https://github.com/flo-renaud/freeipa/pull/56

amore17:
LGTM approved

amore17 (approved):

PR #514, opened 2024-11-06
On F41+, dnf config-manager --set-enabled does not exist

flo-renaud:
Fedora 41 ships with dnf5 and the command
dnf config-manager --set-enabled updates-testing
should be replaced with
dnf config-manager setopt updates-testing.enabled=1


f-trivino (approved):
LGTM! tested locally, worksforme.

Issue #512, opened 2024-11-05
Failure in the playbook ansible/roles/windows/sshd/tasks /main.yml


flo-renaud:
The upstream nightly tests using runAdTests are failing in the step configuring cygwin ssh, in the playbook ansible/roles/windows/sshd/tasks /main.yml.

Example of failure: in [testing_ipa-4.10_latest_selinux] Nightly PR [4121](https://github.com/freeipa-pr-ci2/freeipa/pull/4121)
`test_http_kdc_proxy`: [Details](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/86194dc8-9811-11ef-8e1c-fa163ed415c1/)

The [runner.log](http://freeipa-org-pr-ci.s3-website.eu-central-1.amazonaws.com/jobs/86194dc8-9811-11ef-8e1c-fa163ed415c1/runner.log.gz) shows:
```
2024-11-01 05:31:26,549    DEBUG  TASK [windows/sshd : Check if cygwin sshd service exists] **********************
2024-11-01 05:31:31,195    DEBUG  ok: [ad-root]
2024-11-01 05:31:31,199    DEBUG  
2024-11-01 05:31:31,199    DEBUG  TASK [windows/sshd : Download cygwin installer] ********************************
2024-11-01 05:31:37,909    DEBUG  changed: [ad-root]
2024-11-01 05:31:37,911    DEBUG  
2024-11-01 05:31:37,911    DEBUG  TASK [windows/sshd : Install cygwin OpenSSH] ***********************************
2024-11-01 05:33:52,782    DEBUG  changed: [ad-root] => (item=http://mirror.cs.vt.edu/pub/cygwin/cygwin/)
2024-11-01 05:33:52,787    DEBUG  skipping: [ad-root] => (item=https://mirror.easyname.at/cygwin/) 
2024-11-01 05:33:52,805    DEBUG  fatal: [ad-root]: FAILED! => {}
2024-11-01 05:33:52,805    DEBUG  
2024-11-01 05:33:52,806    DEBUG  MSG:
2024-11-01 05:33:52,806    DEBUG  
2024-11-01 05:33:52,806    DEBUG  The conditional check 'cygwin_result is undefined or cygwin_result.failed' failed. The error was: error while evaluating conditional (cygwin_result is undefined or cygwin_result.failed): 'dict object' has no attribute 'failed'. 'dict object' has no attribute 'failed'
[...]

PR #515, opened 2024-11-06
Cygwin installation: define default value in the loop

flo-renaud:
The playbook installing cygwin openssh contains a loop trying multiple mirrors until one of them succeeds. Ansible changed behavior and the result.failed variable is not set if the iteration succeeded.
Default to false if not set, so that the loop continues if the last iteration failed but is skipped if it succeeded.

Resolves: https://github.com/freeipa/freeipa-pr-ci/issues/512


flo-renaud:
Tested on a personal runner with https://github.com/flo-renaud/freeipa/pull/66

f-trivino (approved):
LGTM!
tested locally, worksforme.

PR #517, opened 2025-03-14
Allow list: fix case for KoRORland

flo-renaud:
The user was added in lowercase but the allow list is case-sensitive.


flo-renaud:
@KoRORland

can you check if your username is correct?

PR #520, opened 2025-06-03
Support modern-webui

duzda:
Due to integration of submodules into freeipa, we need to update and initialize submodules in pipeline as well.
Also ensures the distributable is the same as the release file in pagure, by running autoreconf and make.

### DIFF
--- a/ansible/roles/builder/build/tasks/create_sources.yml
+++ b/ansible/roles/builder/build/tasks/create_sources.yml
@@ -10,6 +10,36 @@
     path: /root/rpmbuild/SOURCES
     state: directory
 
+# We check for Modern WebUI to differentiate between main and older versions
+# as Modern WebUI is not backported to older versions.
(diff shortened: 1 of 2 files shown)


ansible/roles/builder/build/tasks/create_sources.yml
```diff
@@ -10,6 +10,22 @@
     path: /root/rpmbuild/SOURCES
     state: directory
 
+- name: install Modern WebUI dependencies
```
flo-renaud, line 13: Can you just add a comment explaining that the 3 steps (install nodejs, call `npm install` and `npm run build`) are done here in order to create the files for the new webui **outside** of the mock chroot? This way the tar.gz file contains the webui files in `install/freeipa-webui/dist` and they can be directly used in the chroot without any need to install Node.

flo-renaud:
:heavy_check_mark:  New code deployed on our PRCI runners

PR #522, opened 2025-07-02
Add Veronika to our allow list (allowed to trigger CI)

duzda:


flo-renaud:
The change has been deployed on our PRCI runners

PR #529, opened 2026-01-22
Build RPMS with systemd-sysusers configuration

rjeffman:
To build Freeipa with support for systemd-sysusers configuration we need to copy the file 'freipa-sysusers.conf' to the rpm build SOURCES directory, as it is used during build, as a SOURCE, by the specfile.


abbra:
I don't think this change is needed at all. I left my comment in the original IPA pull request with recommendations.

rjeffman:
I agree. Dropping this PR.

PR #535, opened 2026-04-30
Autocleaner: remove boxes for older branches

flo-renaud:
1. When a PR is backported to older branches, the box gets downloaded but is not cleaned up as its definition remains in use for the ipa-4-x branch.
Change strategy and keep only the boxes for the master branch that are actually used.

2. The box name in /var/lib/libvirt/images does not always follow the format
    freeipa-VAGRANTSLASH-{name}_vagrant_box_image_{version}.img
We sometimes have instead
    freeipa-VAGRANTSLASH-{name}_vagrant_box_image_{version}_box.img

Call virsh vol-delete --pool default <name> with both.

### DIFF
--- a/ansible/cleanup_runner_disk.yml
+++ b/ansible/cleanup_runner_disk.yml
@@ -47,4 +47,15 @@
           failed_when:
             res.rc != 0
             and "not found" not in res.stderr
+
+        - name: delete image with name _box.img from libvirt
+          shell: >
+            virsh vol-delete --pool default
+            {{ box_initial_name }}-{{ box_name }}_vagrant_box_image_{{ box_version }}_box.img
+          register: res
+          changed_when: res.rc == 0
+          failed_when:
+            res.rc != 0
+            and "not found" not in res.stderr
+
       when: user_confirmation == "YES" and box_name != "" and box_version != ""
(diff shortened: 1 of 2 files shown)


flo-renaud:
For reviewers:
the PRCI runners have a cron job that regularly cleans up old jobs, old vagrant images etc...
They often run out of disk space because the vagrant images do not get deleted:
- the command is using a wrong volume name
- the images are deleted only if they are not referenced in their respective branch.
For instance if an image is called freeipa-VAGRANTSLASH-ci-master-f43_vagrant_box_image_0.0.2_box.img, its branch is master. The script looks for all the templates referenced in the ipatests/prci_definitions/* files on th emaster branch. It finds
```
        template: &ci-master-latest
          name: freeipa/ci-master-f43
          version: 0.0.2

        template: &ci-master-previous
          name: freeipa/ci-master-f42
          version: 0.0.9

        template: &ci-master-frawhide
          name: freeipa/ci-master-frawhide
          version: 0.10.1
```
If the box matches one of them, it is kept. The issue with this strategy is that it keeps all the images for the older branches.

Deleting an image doesn't have a big impact as it will be downloaded if needed from our prci-automation machine.

duzda (commented):
The box name makes sense, it takes some time but the class `VagrantBox` gives a hint.

There is also ansible job in `ansible/cleanup_runner_disk.yml`, might as well update that one too?

I'm not entirely sure, where the `_box` suffix is coming from.

flo-renaud:
> There is also ansible job in `ansible/cleanup_runner_disk.yml`, might as well update that one too?
> 
Thanks, I missed that one. Patch updated with the same logic in cleanup_runner_disk.yml

duzda (approved):
I'm not entirely sure where the _box suffix is coming from, but I can see that the vagrant-libvirt plugin asserts, that the boxes have the _box suffix, therefore this should indeed work as intended

flo-renaud:
Deployed on PRCI runners

PR #536, opened 2026-05-13
Vagrant images: prepare for fedora 44

flo-renaud:
- Vagrant images: prepare for fedora 44
- Create vagrant box: update the query for selenium download
Update the query for selenium server (the site sometimes exposes the jar
with content_type: 'application/java-archive', and sometimes with 'raw').

### DIFF
--- a/ansible/roles/machine/setup/tasks/install_packages.yml
+++ b/ansible/roles/machine/setup/tasks/install_packages.yml
@@ -158,4 +158,4 @@
         url: "{{ download_metadata.json  | to_json | from_json | json_query(query) }}"
         dest: /opt/selenium.jar
       vars:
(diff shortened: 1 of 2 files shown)


flo-renaud:
Used to create ci-ipa-4-13-f44 v [0.0.1](https://portal.cloud.hashicorp.com/services/vagrant/registries/freeipa/boxes/ci-ipa-4-13-f44/versions/0.0.1) and ci-master-f44 v [0.0.1](https://portal.cloud.hashicorp.com/services/vagrant/registries/freeipa/boxes/ci-master-f44/versions/0.0.1)

duzda (approved):
From a quick look it looks like octet-stream is used for snapshot and stable always uses `java-archive`, on top of that, there is only single .jar file, so the simplification is welcome and more future proof. Thanks.

duzda:
My bad, we simply take the first record, but that renders the removed check useless as well.

flo-renaud:
Deployed on the runners

PR #537, opened 2026-05-15
Add 06shalini to allow list

flo-renaud:
Add Shalini Khandelwal to the list allowed to trigger PRCI


flo-renaud:
Deployed on the runners

